PR #45, opened 2022-03-15
Issue39

### LINKED ISSUE #39: Memory allocation in phenotype proximity
Currently the phenotype proximity workflow uses a distance matrix calculation to simplify the "algorithm". This is now causing overallocation of memory.
Here is my proposed fix:
1. Determine a square grid size by some heuristic (e.g. if P is the largest phenotype-to-phenotype distance considered, use a size larger than or equal to 2P). Play with increasing this to keep the grid tile number down, if this increases the computation speed.
2. Sort the cells by the following order: C1 < C2 if the grid tile of C1 is less than the grid tile of C2 (using lexicographic order on integer pairs).
3. For each grid tile that occurs, compute (i) the square distance matrix between cells in that tile and other such cells, and (ii) the block matrix of distances between cells in that tile and the 4 neighboring grid tiles: E, SE, S, SW.
4. Use these distance values in more or less the same way as they are currently used.

jimmymathews:
This is now implemented, using sklearn.neighbors.BallTree.query_radius. See [here in phenotype_proximity/core.py](https://github.com/nadeemlab/SPT/blob/4a173c10e4d89d79c85e9dfe455373b34d2e658f/spatialprofilingtoolbox/workflows/phenotype_proximity/core.py#L251).

The performance on the tiny test data is about 3 to 7 times slower. I suspect that on the larger datasets this behavior will be completely reversed. In any case, the memory usage was too high before (complete failure), and with the BallTree-based method we will keep this under control.

To provide some assurance that the behavior after the change is correct, I added a new [unit test](https://github.com/nadeemlab/SPT/blob/4a173c10e4d89d79c85e9dfe455373b34d2e658f/tests/unit_tests/test_proximity_core.py#L14) based on a contrived dataset with 10 points equally spaced around a center point, the first 5 belonging to one phenotype, the next 5 belonging to a different phenotype.

Using the unit test, I was able to verify that both the previous behavior and the new behavior are correct (rather than just consistent with each other), at least in the contrived context.

The unit test also uncovered that there is a very slight difference in behavior: BallTree.query_radius is *inclusive* in case of points exactly `r=radius` away.

Closes #39


rvanguri:
One issue that I've noticed with this implementation is that query_radius can include the point being queried on. I'm not sure what the efficient solution to this is other than removing the point being queried on from the list before the tree is built.

jimmymathews:
Got it. Yeah I believe the old method was correct in this case.
So I guess the right fix is to apply a -1 correction for each element of the intersection of the two cell sets.

Issue #85, opened 2022-10-06
GNNs


jimmymathews:
This is graph neural networks modeling and statistics, to be implemented as a new computational-visitor-patterned workflow in SPT. It will start from any dataset available in an ADI-structured database, perform training / model fitting on each available outcome, and send computed features into the database.

May be convenient to complete #83 first as a warmup.

## Acceptance

1. A new workflow is added.
2. The workflow directory contains a module test and at least one unit test.
3. Pass the test suite.
4. The trained model gets saved somewhere with some minimal amount of metadata around it describing the circumstances of its birth.
5. Tried on a major exemplar dataset, results inspected.

CarlinLiao:
I'm maintaining the GNN functionality in an independent repository (on the MSK GitHub, but it doesn't have to be there) that I'm continuing to develop.

We want to merge those features into a sub-sub-directory in the SPT workflows module, ideally while still allowing for pulling additional changes from the original GNN repo, and to update all your docker dependencies etc. to support it. This comes with two main issues

- [Git supports merging two independent histories together,](https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories) but not so much merging one repo into a subdirectory of another repo. [There does exist a tool to support this,](https://github.com/newren/git-filter-repo) but it'd create a weird and niche development dependency.
- The GNN workflow's dependencies are a bit different than the rest of SPT's and I'm not sure how far we can go to isolate them. For example, it requires Python 3.9 instead of 3.10 and would like (but doesn't require) CUDA GPU support.

To do the merge, I have to mess with the directory structure in the original GNN repo, in a way that's less compatible with its independent usage, not to mention the need to add files like core, initializer, and integrator to follow the SPT workflow.

We could just wait until I'm done building out the GNN module before we work on merging it in, but that doesn't allow for parallel testing, and the feature is mostly ready to integrate anyway, barring a few additional metric analysis and visualization add-ons.

(Speaking of, I'm not sure what the intent is to do with the GNN feature. Output some nice graphs and scoring? Identify more important features? etc.)

CarlinLiao:
Perhaps we should reiterate and clarify the scope of the GNN feature:

- Train a model on graphs created from histological structures (i.e., cells only and not tissue), with multiple graphs (ROIs) per slide, against a slide classification of the user's choice 
[...]

Issue #76, opened 2022-09-21
Explicit expression matrix


jimmymathews:
*For final state of criteria/features provided, see the rest of this comment thread.*

There are many usage scenarios where a model or statistic implementation requires a matrix representation of all "features".

Let's create a table, with a record for each cell, with fields:

1. Study identifier
2. Subject identifier
3. Specimen identifier
4. Outcome
5. Coordinate 1
6. Coordinate 2
7. <channel name> (one for each channel)

There are some limitations of this approach:
- Only one outcome.
- Intensity columns and discrete value columns not simultaneously present.

I think the benefit of simplicity outweighs these limitations for most scenarios. That is, e.g., removing the need for the model code to parse columns as outcome columns (there being just the one).

Create a workflow that writes an enormous TSV cells file with all of these fields. As a helper, include a simple JSON lookup of the identifiers pertaining to the first 3 fields (which should normally be occupied by integers): study names, subject names/identifiers, specimen names/identifiers. In case of anonymization only the study names will be non-trivial.

jimmymathews:
This might result in an unmanageable dataframe / file if done as stated.
Let's use multiple dataframes in a nested dictionary, so that columns 1,2,3 are not needlessly repetitious.

jimmymathews:
Probably this can use `spt countsserver cache-expressions-data-array`, as long as we're willing to use a file-mediated method. The file are a very efficient representation.

jimmymathews:
This is being worked on in branch `issue76`. The plan is:

1. Normalize the db access even more. Every use of DatabaseConnectionMaker has some variant of the enter/exit pattern manually implemented; this will be factored into DatabaseConnectionMaker. This is not strictly part of this issue but it is helpful for it, so I'm taking the opportunity.
2. Factor countsserver::CompressedMatrixPuller into the portion that pulls a sparse representation into Python execution environment, and a portion that writes the binary format. The former portion will become a general-use function under workflow::common.
3. Add the expression matrix functionality by doing the sparse representation pull plus some dataframe stuff. Make sure this is available as dataframe, dictionary, or file write. The file write should probably be normally used via a new exposed CLI command, which doubles as a documentation vehicle for the new functionality as a whole.
4. Add a strict test of the expression pulling, testing that the resulting dataframe is exactly as expected. Possibly add a few finer-grained tests (for the future, in case of deliberate changes, to test the deliberate changes).

jimmymathews:
(1) above is done.

[...]

PR #93, opened 2022-11-02
Add cg-gnn pipeline

CarlinLiao:
Per #85, all feature requests have been resolved and we're ready to begin work integrating the CG-GNN pipeline into SPT.


CarlinLiao:
I want to take a moment to note that I've carefully preserved and merged in the commit history for all of histocartography, hact-net, and patho-quant-explainer instead of directly copying their code without line-by-line attribution.

jimmymathews:
We should check their licenses to make sure whatever we're doing is OK.

jimmymathews:
Just noting: This branch seems to add about 200MB to the repository.

jimmymathews:
I'm counting about 5000 new lines of python code. Does this match your expectation? Do you know roughly how much of this is from the other repos?

For reference the SPT codebase has about 7000 lines of python code.

jimmymathews:
I would like to request that the identifiers be systematically expanded, for example

- `thresh -> threshold`
- `val -> validation`
- `prc -> percent`
- `bboxes -> bounding_boxes`
- `k -> ...`
- `i_specimen -> ...`
- ...

I don't think this expansion is necessarily desireable for the external code that has been incorporated, for all the normal reasons involving external code. But is there a way to mostly use the external code with the normal library pattern instead? That is, by Installing their packages from source releases.

In cases where some small tweak to an external repo's source file is necessary, I'd be open to using the "diff patch" system. So only the patch would be committed to our repository.

CarlinLiao:
The source repos are AGPL-3 or MIT licensed so we're good on that front.

It is big. This is a heavily modified version of three different repos, only one of which is available on pypi and supported (and even that is more in theory than in practice given their lack of response to issues I've opened). I went in and updated what we've kept to be compatible with more modern versions of the dependencies used, so there's no guarantee that future updates of the source will be compatible with any updates they put out. I'm sure we could make it work, but I think it'd be more trouble than it's worth.

I can go in and expand the variable names no problem, though.

jimmymathews:
I see, that makes sense. I shouldn't have emphasized keeping up to date, because I think the more important thing is functional isolation. I'd like to be able to use their code as a library, so we don't ourselves implicitly accept the burden of maintenance on a large codebase that we haven't reviewed/written.

Since you're saying they aren't on PyPI, maybe instead we can use commit hashes from the git repositories as a proxy for release version, then use diff and patch in order to keep straight exactly what changes we have introduced (hopefully this is only a handful of lines? How many are we talking, a few dozen?)

jimmymathews:
The part of our "build" process is:

1. git pull a specific commit
2. apply patches

CarlinLiao:
As we discussed in person, we're going to change this integration by separating out the CG-GNN functionality into a [standalone library](https://github.com/CarlinLiao/cg-gnns) published on PyPI that SPT will use as a dependency.

CarlinLiao:
I expected publishing to PyPI to be trivial but oh how wrong I was.

The root of the problem is that the dependence on DGL and pytorch are weird because
1. their downloads are maintained off of pypi, maintained by their own development teams
2. the version installed varies depending on which version of CUDA, if any, the local environment has

After a lot of trial and error and reviewing how my own development environment has works, I assume this is usually resolved by requiring the user to install pytorch and DGL themselves and just having the library error if they're not present. Working on this now.

CarlinLiao:
[Now on PyPI](https://pypi.org/project/cg-gnns/)

CarlinLiao:
I've overwritten the prior attempt at directly merging `cg-gnns` into SPT with a simple commit that just adds a folder to the workflows directory for us to work the `cg-gnns` code into.

To get it to work we'll either need to have much stricter requirements for the SPT conda env or containerize this workflow. It requires
* `python=3.9`
* `torch`
* `dgl`
* `cg-gnns` on PyPI, of course

If a GPU that supports CUDA is available on the source machine it should also install cudatoolkit and the right cuda-supported versions of torch and dgl.

jimmymathews:
Great!
Yes this seems to fit the module pattern of this repository. Each module gets its own list of dependencies, and its own Dockerfile for a separate container definition. It should be no trouble specifying the installation of prerequisites in the Dockerfile. The question is just whether to include pytorch etc. in `workflow` module or create a new `gnn`. The argument for the latter is that the dependencies are large, and other workflows won't necessarily need them.

CarlinLiao:
While putting this together I ended up reading some discussion from PyPI about what versions of packages they choose to host. Long story short if they actually paid for their hosting services it would cost millions, and that would be much, much greater if they hosted packages suited for different GPUs in addition to different software versions and CPU types. GPU packages also tend to be much larger, so we probably do want to containerize.

This may also be the only workflow that wants a GPU, so it ought to be a better use of hosting funds as well.

CarlinLiao:
Here's my best guess at what the `cg-gnn` module should look like. I wasn't sure what Makefile pattern to follow since most if not all of the other modules spin up a new server for the database, counts, or what have you, whereas use of this module doesn't. It's only supposed to query an existing SPT postgreSQL server do some processing in its Docker container, output logs, and save stuff to file. (Ultimately we'd probably want to break up the `cg-gnn` functionality into separate commands, but for now this just runs everything in one shot.) I guess the best comparison would be like an isolated version of the `spt db create-schema` script.

I also wasn't sure where to put the requirements for this isolated Docker container.

When you have time, take a look and let me know how I should set up the Makefile, command line argument format, Dockerfile, and whatever else is necessary to get it up to speed.

jimmymathews:
A quick note: You say that the other modules "spin up a new server for the database" and your module "is only supposed to query an existing SPT postgreSQL server", but I think that the pattern is the same not different. In order to test the workflow code, you don't want to query an existing (production) SPT postgres server but rather the test-data-preloaded database, exactly as the other modules do.

CarlinLiao:
Fair point, but don't the other modules like countsserver spin up their own server even in normal usage outside of testing?

CarlinLiao:
Quick update here, still getting to grips with Docker. I've learned enough about compose to formulate this question: how do we want to use and test the cg-gnn pipeline? Here's what I'm thinking:

## A proposal for how it should work

Calling `spt cg-gnn [args]` will create a new, isolated environment for cg-gnn to be installed, and then runs that that instance on the args being provided to SPT before disposing of the env. (Alternatively, the first call instantiated the cg-gnn env and persists it for future runs.)

This would fully isolate `cg-gnn` dependencies from SPT dependencies, which is a big benefit since not all SPT users will utilize the neural net pipeline and PyTorch, etc. take a _lot_ of space as we've discussed when it comes to Saad's remote.

The isolated env can be as simple as a Python virtualenv, conda env, or an entire Docker instance. A potential downside of the latter is how it'd require Docker to be installed and usable on the machine you're using SPT on, although given that all the other parts of SPT minus the command line interface requires SPT I doubt this is a significant restriction. This is also true to a lesser extent with conda.

Another downside is the need to create and maintain two envs, one of which is only instantiated at runtime. This introduces more complexity and slows down the call process.


## How to test it

I think the most applicable pattern to follow would be the compose and Dockerfiles assembled for the `spatialprofilingtoolbox/workflow` pipeline. I can just straight copy the `compose.yml` for the former with only the test-db service, while the latter can just be `building/Dockerfile.base` with no appending unless we go the conda or nested Docker instance route.

(By the way, the Dockerfile.base and Dockerfile.append nomenclature threw me for a loop even after reading the Docker quickstart documentation. After further reading I think this is a formula you found convenient for your Makefiles rather than a standard feature of Docker?)

CarlinLiao:
Actually, since torch and DGL can't be installed via requirements.txt since they're hardware-dependent, we could avoid the whole nested-env nonsense by just including `cg-gnn` as a dependency in SPT (and by extension its non-deep-learning dependencies) and letting it notify the user that they need to install torch and DGL themselves to use the `cg-gnn` module.

(By the way, this only works as long as SPT is limited to py3.9, which it is, since as of now `cg-gnn` still breaks in py3.10.)

CarlinLiao:
These changes are pursuant to the last proposed solution of having optional dependencies that build if you enable the cg-gnn feature, as with the apiserver, db, countsserver, etc.

I haven't checked if the CLI is picking up the new cg-gnn command, but the new Dockerfile.append has been incorporated into the `make test` suite. That said, the image I've specified fails to build because the `cg-gnn` package depends on mlflow, which depends on `shap`, which fails to install because it's expecting CUDA even though that should be an optional dependency for it.

My solution is to go back into cg-gnn and remove mlflow as a dependency, although I'm concerned that this might require reworking a lot of the model saving and loading processes. Luckily mlflow isn't a dependency of histocartography, so when they fix their pip package and `cg-gnn` can depend on that instead of maintaining my own version of their module nothing will break.

CarlinLiao:
I am less clear than before. The solution I'm working on doesn't rely on any docker images at all outside of testing, or at least I thought it didn't. My understanding was that calling spt commands ran them locally instead of creating a new Docker image, and that was just a convenience done for testing and validation, but it sounds like that's not the case?

Either way, for now I'm just following the pattern set by the other modules in `spatialprofilingtoolbox/` so it should be consistent, I just appear to have misunderstood what that does.

Have you disambiguated the SPT testing process from the use process? E.g., the compose.yml files seem to target testing the submodules, but they seem unnecessary for runtime, but I don't see any compose.yml files that target runtime.

CarlinLiao:
I've updated the pip release of `cg-gnn` to not use mlflow, so the docker image build has moved onto something different to fail on.

The root of it is the preference but not requirement for CUDA. Having a CUDA driver for a compatible GPU speeds up training time by orders of magnitude so it's almost a requirement. The trouble is, installing it in a Docker container is somewhat complicated and I can't figure out how to do it elegantly. The exact way to do so is also highly dependent on both the Linux version (though it looks like we're sticking with Alpine) and the actual hardware being used (ideally, should be conditional on if there's a CUDA-enabled GPU or not).

jimmymathews:
Your understanding is very close. spt commands do indeed run "locally" provided you have installed them locally. I often do this for basic `spt db` stuff.
However, this repo does some of the additional legwork of pre-installing these commands together with various dependencies into various Docker images.
These Docker images are then tested in various test environments, that is, the Docker "compositions" defined via those `compose.yaml` files.
The Docker images are also used in "production", notably in the spt-dashboard application (located in a separate repository).
This setup allows the test environments to very closely simulate the environment that the services (individual Docker containers templated on a given Docker image) encounter at production runtime.
In short, the same Docker images that are tested are then deployed in production.

jimmymathews:
For performance purposes, and also because of Dockerlessness, on the HPC I've been using the `workflow` module from a direct installation, e.g. in a conda environment:

```sh
pip install spatialprofilingtoolbox[workflow]
```
Then directly doing `spt workflow configure` etc.

It sounds like this will need to be the pattern for `cg-gnn` as well, due to things like dependence on GPUs.

jimmymathews:
The fact that the CUDA requirement is a hard requirement only in real runs (due to speed) probably means that we can do: The Dockerized version of the `cg-gnn` module built for the purpose of testing functionality, while the real usage pattern is Dockerless with direct installs.
This is similar to the situation for `db`. The `db` Docker image, containing a Postgres database, is never meant for production, but it is very convenient to have a test version of it available (1) for testing `db` modules functionality, and (2) for mocking data for other modules' testing environments.

jimmymathews:
A potentially clarifying note: I noticed that many people seem to write Docker compositions (`compose.yaml`) in such a way that the Docker images that are invoked are also *built* at compose-time. That's a little weird to me and seems like mixing up runtime considerations (docker compose) and compile-time considerations (docker build). It's possible that you thought that spt is doing this, and in this case you might have been led to think that the Dockerization is only happening in test environments?
The `docker build` commands are invoked by Make at the top level: `make build-docker-images`.

CarlinLiao:
We're going to have a hard time running the cg-gnn pipeline as intended because we lack a test machine that supports both CUDA and Docker.

If I understand correctly, you're suggesting we test two different versions of this workflow
1. a Dockerized version that assumes no CUDA
2. a locally running version that we can test on the NadeemLab HPC that supports CUDA

1 will work for meeting our integration requirements but for production we need have a conditional Docker build that builds differently depending on whether CUDA is available, and the difference isn't trivial as far as I can tell.

2 I think is the more realistic use case, but won't be runnable in isolation since the pipeline needs a database to query. For my own local tests of the pip package I assume the scstudies server is up and just use that, so that would just be a matter of copying the test specification over.

CarlinLiao:
Does the production spt-dashboard use the compose.yml files in this repository, or does it have its own?

CarlinLiao:
This latest Dockerfile supports CUDA 11.6 but still builds on my CUDA-less Macbook. My guess is it'll error when I try to spin up an instance, but at least it builds.

CarlinLiao:
I edited the cggnn compose.yml to include docker images for both the test db and cggnn. I tried to run it on my machine but it fails, I think because the docker image for cggnn, based on Ubuntu with torch, isn't built for amd64/Apple Silicon. Maybe it'll work on your machine since it's Intel?

jimmymathews:
Sorry I forgot to answer this: I don't think it should matter? I thought Docker's whole thing is that it handles virtualization of OS and hardware... Maybe this is just Docker's fault though, insufficient support for different new CPU architectures.

CarlinLiao:
It makes sense to me. You can virtualize an operating system but you can't virtualize a completely different system architecture and a different processing unit. Docker supports emulation of different hardware but that can only go so far.

I don't think Docker would want to support running x86 applications on ARM and vice versa because there would be such a significant performance hit either.

CarlinLiao:
[Docker doc on multiplatform images.](https://docs.docker.com/build/building/multi-platform/) The `from` clause pulls the image for the relevant machine architecture, if there's one available.

[The torch image on docker hub only has a linux/amd64 variant, so no ARM support.](https://hub.docker.com/r/pytorch/pytorch/tags)

jimmymathews:
I see, yeah there are limitations to cross-platform stuff.

CarlinLiao:
@wooferclaw Have you been able to test the Docker pipeline on your system?

CarlinLiao:
I've finally been able to secure a compatible VM to test this on. Both currently written tests (image builds, pipeline runs without error) pass. Is this ready to merge?

Makefile
```diff
@@ -22,7 +22,7 @@ help:
 >@${MESSAGE} print '      make test'
 >@${MESSAGE} print '          Do unit and module tests.'
 >@${MESSAGE} print ' '
->@${MESSAGE} print '      make test-[apiserver | countsserver | db | workflow | ]'
+>@${MESSAGE} print '      make test-[apiserver | cggnn | countsserver | db | workflow | ]'
```
CarlinLiao, line 25: I'm a bit troubled by this help text ending with a |

Also, this help text is wrong as we discussed earlier. This help should also be included in the maintenance.md doc.

jimmymathews: Yes you are right!

CarlinLiao: I added the new help text in #103

jimmymathews:
The CLI entrypoint autodiscovery of cggnn's "run" command works at the level of tab completion, but the spt.py entrypoint itself seems to be failing to discover the cggnn module. Trying `spt cggnn run --help` just prints SPT version, as though the command was not recognized.

I'm tracking it down...

jimmymathews:
I spotted it, subpackage `spatialprofilingtoolbox.cggnn` was not declared in `pyproject.toml`.
Now `spt cggnn run --help` does indeed run the new `run.py` script.
The new issue is that there is no error guard against external package `cggnn` being installed. (When it is not installed, the help text cannot even print).

jimmymathews:
I added the dependency hint (in case `cggnn` isn't installed) to the script's import, copying what was done in `db` etc. You might not like that pattern, conditional import... Feel free to suggest an alternative.
We now get this hint printing:
```sh
$ spt cggnn run --help

Got a module not found error.
Did you install the required extras with:
    pip install "spatialprofilingtoolbox[cggnn]"
?

...
```

jimmymathews:
By the way, I'm surprised the test was passing given that `spt cggnn run` wasn't calling up the `run.py` script...

jimmymathews:
Previously (before I made any commits on this branch), I was getting a pass for the `run` test, but now it fails because it is actually running.
My error is that the `spt-db-testing` server isn't up in time to respond to queries. This happened in all other modules' tests as well, which is why I wrote the "poll for container readiness" utility. In your Makefile you changed the setup line that calls this to wait for `spt-cggnn-testing`, but it needs to wait for `spt-db-testing`.
I think you also didn't mean to spin up the service `testing-cggnn` in your `compose.yaml`... You are running tests directly in a container based on the cggnn image, and it is not meant to be run as a server as far as I can tell (there is no `CMD` line in the Dockerfile).
I'm modifying the Makefile and compose.yaml accordingly.

jimmymathews:
(Side note: On my machine the additional disk usage by Docker corresponding to this branch is about 12-14 GB, mostly the pytorch image. And the download and build took about 15 minutes.)

jimmymathews:
I now got my first actual error, I think:

```sh
  File "/opt/conda/lib/python3.9/site-packages/spatialprofilingtoolbox/cggnn/scripts/run.py", line 141, in <module>
    run_pipeline(args.measurement_study,
  File "/opt/conda/lib/python3.9/site-packages/cggnn/run_all.py", line 36, in run_pipeline
    df_cell, df_label, label_to_result = spt_to_dataframes(
  File "/opt/conda/lib/python3.9/site-packages/cggnn/spt_to_df.py", line 211, in spt_to_dataframes
    df_cells = _create_cell_df(_get_targets(conn, measurement_study),
  File "/opt/conda/lib/python3.9/site-packages/cggnn/spt_to_df.py", line 144, in _create_cell_df
    columns: List[Union[int, str]] = list(range(df_targets['target'].min(),
TypeError: 'float' object cannot be interpreted as an integer
```

jimmymathews:
(By the way, if I understand it correctly the functionality of `spt_to_df.py` will be provided by #76 soon)

CarlinLiao:
Not sure about the specific error, but I suspect it has something to do with cggnn not being updated to match the new SPT data schema.

jimmymathews:
Shall we merge? I think the functionality in this PR seems OK, the test failure is due to the (currently external) cg-gnn package.

Issue #84, opened 2022-10-06
Linting and IDE setup


jimmymathews:
Use an IDE, e.g. PyCharm or Visual Studio Code, to review and address compliance with conventions and to autodetect low-level bugs in the SPT codebase. After fixes the test suite should be run frequently to ensure continuity of behavior. Possibly, do any background work needed to make usage of the IDE the primary recommended mechanism for development.

## Acceptance

1. Commit a file containing the list of issues, plus "addressed" or "not addressed", to an issue branch.
2. If an IDE will be recommended, add a blurb about getting started with it to the developers' documentation

## Notes

A linter was not used very often during development, and neither was an IDE. This has most likely resulted in the occurrence of many small style errors, and perhaps larger real errors as well.

CarlinLiao:
Working on this now. I'd add as development dependencies

* autopep8
* mypy
* pycodestyle
* pylint

to enforce this via pyproject. I think that this should make it PyCharm and VS Code independent, but I've only used VS Code for linting.

How involved would you like this resolution to be? #86 requests a full codebase review whereas linting can be as simple as
1. Applying whatever autopep8 does to each .py file
2. Manually fixing every error raised by a linter that can't be handled automatically, which includes but isn't limited to
   * Removing unused imports
   * Reordering imports to follow recommended style (standard, third-party, local)
   * Lines that are too long
   * Lack of type hints on functions
   * Missing docstrings on every method and class
   * Vague function names
3. (This isn't actually linting but) replacing full library imports with only importing the functions needed from the library

jimmymathews:
This sounds perfect.
My work on #94 is touching a lot of the codebase, so maybe wait to do your initial pass via the automated tooling until `issue94` is merged.
You can update [maintenance.md](https://github.com/nadeemlab/SPT/blob/main/docs/maintenance.md) with instructions for getting the linting and IDE set up. Feel free to be very verbose there, I'm not too familiar with the formal conventions.

CarlinLiao:
After discussion, we decided to
* spin out docstrings into a separate issue
* spin out type hints into a separate issue
* variable renames we'll consider to be out of scope
* this issue will be resolved by accomplishing the rest of the simple manual fixes in addition to the autopep8 changes

CarlinLiao:
List of linter complaints that I'm not going to fix for this issue because they change functionality in addition to the above:
* (pylint)
   * prefer f-strings to `%` string formatting
   * use lazy `%` formatting in logging functions
[...]

PR #101, opened 2022-12-14
applied autopep8 and did a bit more styling

CarlinLiao:
This PR is pursuant to #84. See discussion there.


CarlinLiao:
All tests pass in `main` but not this new branch, even though this branch shouldn't have any functionality changes. That's concerning...

CarlinLiao:
Haven't diagnosed why yet, but my first suspicion would be issues arising from autopep8 automatically reordering inputs you placed in the middle of Python files to the top.

jimmymathews:
A major reason it can be difficult to track down the build errors is that sometimes the real error is at an upstream step that is not caught properly.
In this case it is the docker image build of the pre-loaded data image:

```
Building test-data-loaded spt-db image (1) ...
...
    from spatialprofilingtoolbox import get_workflow_names
ImportError: cannot import name 'get_workflow_names' from 'spatialprofilingtoolbox'
```

This step didn't send a failure exit code up the chain for some reason, a bug in the Makefile.

jimmymathews:
Still, you can find these errors by doing `make clean && make test VERBOSE=1` and just checking the logs.

spatialprofilingtoolbox/workflow/scripts/configure.py
```diff
@@ -1,48 +1,52 @@
...
 from os.path import join
 from os.path import abspath
 from os.path import expanduser
-from os import getcwd
+import stat
 import importlib.resources
 
-import spatialprofilingtoolbox
 from spatialprofilingtoolbox import get_workflow
 from spatialprofilingtoolbox import get_workflow_names
```
jimmymathews, line 13: The error we got was here.
The only difference is `import spatialprofilingtoolbox` above is gone!
This matches the experimental behavior I thought I observed in the past, that led me to systematically include `import <package>` before getting anything from it.

spatialprofilingtoolbox/workflow/dataset_designs/multiplexed_imaging/halo_cell_metadata_design.py
```diff
@@ -16,17 +13,19 @@ class HALOCellMetadataDesign:
     This class provides the schema necessary to interpret cell metadata manifests
     exported from the HALO software.
     """
+
     def __init__(self,
-        elementary_phenotypes_file: str=None,
-        compartments_file: str=None,
-        **kwargs,
-    ):
```
jimmymathews, line 23: Was this deliberate? Seems like the previous indenting method here was consistent with the rest of the code.

jimmymathews: Oh I think I see the logic, it's matching the indent of the first open parenthesis?

spatialprofilingtoolbox/workflow/scripts/configure.py
jimmymathews: It seems the problem was in spatialprofilingtoolbox's __init__.py, now fixed.

spatialprofilingtoolbox/workflow/dataset_designs/multiplexed_imaging/halo_cell_metadata_design.py
CarlinLiao: Yes. If `self` was also on its own line instead of with the first open paren, it would follow the original indenting format. It's easy to forget that `self` is also an input arg.

jimmymathews (approved):

Issue #103, opened 2022-12-20
Migrate test and build scripts to separate directories


CarlinLiao:
See #86. Best practice is not to co-mingle the test, build, and deployment infrastructure with the Python files going into the wheel.

# Acceptance

* Only the files going into the Python package wheel are contained in `spatialprofilingtoolbox/`
* Test scripts and support infrastructure are fully contained in `test/`
* Build scripts, makefiles, and other support infrastructure are fully contained in `building/`

CarlinLiao:
I want to rename `building/` to `build/` but it's already there so I guess...

CarlinLiao:
Things I'm a bit confused about how to place:
* db/data_model. This is a bunch of SQL files and I can't dichotomize if they're used for wheel, build, or deployment. I don't think they go into the wheel?
* Ditto with .jinja files

CarlinLiao:
I was so confused reading these Makefiles because I didn't realize what `.RECIPEPREFIX` did. I'd prefer more whitespace but I'll follow your convention here.

CarlinLiao:
Slowly working through this... I think I have a good handle on how the makefiles behave now, and I'm getting a better grasp of what things are used (only?) during build. To solve this issue, we're to take all build scripts out of the main package directory, but identifying what's a build script and what isn't is suprisingly non-trivial.

From my view, here are the ambiguous cases:
* `entry_point` Python scripts are used only during build time to fill out the completions shell script template. Even so, it's dependent on anther `spatialprofilingtoolbox` module so it's not cleanly extractable. Resolving this might require review of pythonic best practice for handling python libraries that are meant to be run as their own independent CLI outside of a python CLI.
* `db/data_model` contains an `__init__.py` even though it only contains SQL files. As far as I can tell, these SQL files are only used at build time to coalesce them into a single `schema.sql` file. I think they should be moved to an outer `build` directory but maybe they get used during runtime as well?

What is the difference between build and runtime for SPT anyway? Running SPT is essentially for the purpose of spinning up (or "building") an SPT server right? So we build stuff to prepare for SPT to build its server. That's kinda confusing.

CarlinLiao:
There's also the related matter of `schema.sql` being created in `db/data_model`, only to be mixed in to `initalize_schema.sql`. For now, I'm leaving `data_model` in the source directory and moving the location where `schema.sql` is built into the build directory, but we could clarify the structure here.

CarlinLiao:
[...]

PR #104, opened 2022-12-31
Migrate test and build scripts to top level directories

CarlinLiao:
See discussion in #103


CarlinLiao:
State of PR as of 2022 (haha)
* building works, with the exception of `apiserver`, specifically the `COPY ./app app` line in its Dockerfile due to the separation of source (app) and build directories (where the Dockerfile is)
* cleaning works
* testing commands have yet to be updated
* some discussion about casting a larger net for ambiguously build- or runtime-related files in `db/data_model` and `entry_point` can be found in #103

CarlinLiao:
Something else that's been tripping me up: inconsistent use of a file's absolute vs. relative location in the makefile. I was debugging why a rule wasn't found by `make` for a while until I realized the rule was defined for the file's _relative_ location when I was using the variable for its _absolute_ location.

CarlinLiao:
Ok, with the latest commit we're almost there. Dry running `make test -n` produces no errors, but I'm still afraid to run it live because I know for a fact that there'll be at least two runtime errors:
1. Issue with building the `app` Docker container as discussed above
2. The mount point provided to the unit and module tests needs to be changed. The Docker images used for testing are built in the `build/` directory, but the test files themselves are in `test/` and the files they're testing are in `spatialprofilingtoolbox/` so this is going to be complicated.

CarlinLiao:
I've run `make test` and now know exactly what's wrong... more or less. Current state (need help!):
1. apiserver Dockerfile (see above)
2. (NEW) `data-loaded-image-` containers fail to build because of `No space left on device` complaints. My laptop has >100 GB of HDD space left, so it must be something else causing it. Am I copying the wrong directory over now, or is this a local issue?
3. Unit and module tests in the Docker images are trying to run from the build folder and not the test folder, but I can't find where they're being referenced so I can change the path. How exactly does this work? Are the test shell and python files
   a. copied or mounted to the container when they're first built, or
   b. provided directly at runtime?

CarlinLiao:
Well, logging in didn't fix it. I'm getting
```
docker: Error response from daemon: pull access denied for nadeemlab-development/spt-development, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```
Is the issue that
1. I don't have access to the development things I want to pull from
2. The name has been compiled wrong
Maybe problems 2 and 3 from earlier are related. The test images are pulling from remote docker images that are reliant on the old directory structure, but trying to run on the new directory structure, so it's confused?

jimmymathews:
The repository nadeemlab-development/spt-development is local only (for development). It doesn't exist to be pulled.

CarlinLiao:
Why is it trying to pull it from remote then? Is it not being created locally and Docker's falling back to web?

Makefile
```diff
@@ -36,26 +38,33 @@ help:
...
 COMPLETIONS_DIRECTORY := ${PWD}/${PACKAGE_NAME}/entry_point
 DB_DIRECTORY := ${PWD}/${PACKAGE_NAME}/db
 WORKFLOW_DIRECTORY := ${PWD}/${PACKAGE_NAME}/workflow
 PACKAGE_DIRECTORY := ${PWD}/${PACKAGE_NAME}
+PACKAGE_DIRECTORY_RELATIVE := ${PACKAGE_NAME}
+COMPLETIONS_BUILD_DIRECTORY := ${BUILD_LOCATION}/entry_point
+DB_BUILD_DIRECTORY := ${BUILD_LOCATION}/db
+WORKFLOW_BUILD_DIRECTORY := ${BUILD_LOCATION}/workflow
+PACKAGE_BUILD_DIRECTORY := ${BUILD_LOCATION}
 export
```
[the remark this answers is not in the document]
jimmymathews: I believe export in Make language makes these variables available in submakes (`${MAKE}`). I'm not sure if it matters when it is called.

jimmymathews:
Regarding absolute vs relative file paths in Makefile: What is an example of this?
I believe make -- in order to give special treatment to targets that are actual files -- requires you to use a target name that is the relative path of the file, relative to the directory in which make is run. I tried to make every target that is a file reference have this format, if not this is a mistake.
However many targets aren't files. Maybe this is what was confusing?

jimmymathews:
Yeah that is the most likely reason!

CarlinLiao:
Interestingly enough I failed to reproduce the error after another clean and rerun. Ran into the port-already-in-use error again though and had to manually kill the postgresql instance on 5432.

jimmymathews:
If I understand the issue with `apiserver` build correctly, we just need to change the `COPY` line to use the appropriate relative path. Perhaps
```
COPY ../spatialprofilingtoolbox/apiserver/app/ /app
```
depending on how you have setup the build directory, I have to look more carefully.
For this to work though we have to be sure that the Docker build context includes this directory. Maybe better to copy in `spatialprofilingtoolbox/apiserver/app/` just for building.

Makefile
```diff
@@ -3,8 +3,11 @@ MAKEFLAGS += --warn-undefined-variables
 MAKEFLAGS += --no-builtin-rules
 
 PYTHON := python
-BUILD_SCRIPTS_LOCATION := ${PWD}/building
-BUILD_SCRIPTS_LOCATION_RELATIVE := building
```
CarlinLiao, line 7: This is an example of the absolute vs. relative location thing. I don't think both are necessary as long as we're consistent about it throughout the makefile. If we use relative in one instance but absolute in another, the prerequisite rule doesn't get picked up and it errors.

jimmymathews: Hm it looks like BUILD_SCRIPTS_LOCATION_RELATIVE is vestigial, it's not used.

CarlinLiao:
Re: ancestor directories in Dockerfiles, I believe that's the case. I tried a simple edit like that and if I recall correctly it didn't work, likely because that build context wasn't passed over.

CarlinLiao:
Current state of this branch. Missing pieces:
* `apiserver` Docker build issue as noted above
* Tests fail to run because they're not found in the directories they're supposed to (and definitely are) in, likely due to a Docker-context-passing issue instead of the files not being where they are. Note that the test files are referred to by their absolute location and not their relative location---initially I had them referred to by relative location but then some prerequisite rules failed to get picked up and run.
```
cliao@LTDV0213 SPT % gmake clean                                                                            
Checking that Docker daemon is running ......................................... Running.       (1s)      
Running docker compose rm (remove) ............................................. Down.          (1s)      
cliao@LTDV0213 SPT % gmake test
Creating venv for basic package introspection .................................. Created.       (2s)      
Creating spt CLI completions script ............................................ Created.       (3s)      
Building development image ..................................................... Built.         (3s)      
Building apiserver Dockerfile .................................................. Built.         (0s)      
Building countsserver Dockerfile ............................................... Built.         (0s)      
Building db Dockerfile ......................................................... Built.         (0s)      
Building workflow Dockerfile ................................................... Built.         (0s)      
Checking for Docker credentials in ~/.docker/config.json ....................... Found.         (1s)      
Building Docker image nadeemlab/spt-db ......................................... Built.         (7s)      
Building test-data-loaded spt-db image (1) ..................................... Built.         (15s)     
Building test-data-loaded spt-db image (1and2) ................................. Built.         (25s)     
Building Docker image nadeemlab/spt-apiserver .................................. Build failed.  (7s)      
Building Docker image nadeemlab/spt-countsserver ............................... Built.         (7s)      
Building Docker image nadeemlab/spt-workflow ................................... Built.         (6s)      
Running docker compose rm (remove) ............................................. Down.          (1s)      
apiserver (setup testing environment) .......................................... Not setup.     (0s)      
  /Users/cliao/projects/SPT/test/apiserver/API internal basic database accessor ... Failed.        (1s)      
apiserver (teardown testing environment) ....................................... Down.          (0s)      
countsserver (setup testing environment) ....................................... Setup.         (2s)      
  /Users/cliao/projects/SPT/test/countsserver/binary expression viewer ......... Failed.        (1s)      
countsserver (teardown testing environment) .................................... Down.          (0s)      
db (setup testing environment) ................................................. Setup.         (2s)      
  /Users/cliao/projects/SPT/test/db/drop recreate database constraints ......... Failed.        (1s)      
  /Users/cliao/projects/SPT/test/db/guess channels from object files ........... Failed.        (0s)      
  /Users/cliao/projects/SPT/test/db/shapefile polygon extraction ............... Failed.        (1s)      
db (teardown testing environment) .............................................. Down.          (1s)      
workflow (setup testing environment) ........................................... Setup.         (2s)      
  /Users/cliao/projects/SPT/test/workflow/centroid pulling ..................... Failed.        (1s)      
  /Users/cliao/projects/SPT/test/workflow/feature matrix extraction ............ Failed.        (0s)      
  /Users/cliao/projects/SPT/test/workflow/outcomes pulling ..................... Failed.        (0s)      
  /Users/cliao/projects/SPT/test/workflow/sample stratification ................ Failed.        (1s)      
  /Users/cliao/projects/SPT/test/workflow/signature cell set subsetting ........ Failed.        (0s)      
workflow (teardown testing environment) ........................................ Down.          (0s)      
apiserver (setup testing environment) .......................................... Not setup.     (1s)      
  /Users/cliao/projects/SPT/test/apiserver/single API route .................... Failed.        (1s)      
  /Users/cliao/projects/SPT/test/apiserver/study summary retrieval ............. Failed.        (1s)      
  /Users/cliao/projects/SPT/test/apiserver/counts query delegation edge cases ... Failed.        (0s)      
apiserver (teardown testing environment) ....................................... Down.          (0s)      
countsserver (setup testing environment) ....................................... Setup.         (3s)      
  /Users/cliao/projects/SPT/test/countsserver/class counts cohoused datasets ... Failed.        (0s)      
  /Users/cliao/projects/SPT/test/countsserver/expression data caching .......... Failed.        (1s)      
  /Users/cliao/projects/SPT/test/countsserver/edge cases few markers ........... Failed.        (1s)      
  /Users/cliao/projects/SPT/test/countsserver/single signature count query ..... Failed.        (0s)      
countsserver (teardown testing environment) .................................... Down.          (1s)      
db (setup testing environment) ................................................. Setup.         (2s)      
  /Users/cliao/projects/SPT/test/db/basic health of database ................... Failed.        (0s)      
  /Users/cliao/projects/SPT/test/db/record counts cohoused datasets ............ Failed.        (1s)      
db (teardown testing environment) .............................................. Down.          (0s)      
workflow (setup testing environment) ........................................... Setup.         (2s)      
  /Users/cliao/projects/SPT/test/workflow/proximity pipeline ................... Failed.        (0s)      
workflow (teardown testing environment) ........................................ Down.          (0s)  
```

Makefile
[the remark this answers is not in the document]
CarlinLiao: [That appears to be correct per the manual.](https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html) So we can move the export statement to the beginning and block out these variable definitions I take it?

Makefile
CarlinLiao: I see. I've definitely switched to using it a lot in this branch. On top of that, all handwritten file references are relative (because, of course, they have to be). I'd prefer to make every location reference relative instead of absolute for consistency's sake, and minimize use of `${PWD}` as much as possible.

The long grep command a couple dozen lines below this, and the rules it sets up, might not be compatible with this though.

CarlinLiao:
After merging `cggnn` into `main` and then this PR, no new problems have developed. The `cggnn` Docker image built without issue, and its tests fail but not for any new emergent reason not previously documented.

CarlinLiao:
The latest commit points the Docker images to the correct location of the test files, so now half the tests pass. I'm debugging why the rests of the test are failing, and my guess is that it's because the test file paths aren't the only ones that need updating.

The first one I've isolated, the `db` test `drop recreate database constraints`, fails because it can't find `/mount_sources/db/.spt_db.config.container`. This led me to an interesting observation:
1. The `db` Dockerfile has a line that copies `.spt_db.config.container` into the working directory of the container, `COPY .spt_db.config.container ./`
2. However, the version of the file that's actually used is the original, mounted onto the Docker image via `docker run ... --mount type=bind,src=${PACKAGE_DIRECTORY},dst=/mount_sources` and then `test_cmd="cd /mount_sources/db/; $$executor $$test_script`

Of these two, which is the intended behavior?
* If 1, we'll need to edit how the test is called by the Makefile recipe, or edit the copy command to move it directly to `mount_sources/db/`
* If 2, I can delete the COPY line from the `db` Makefile and manually copy `.spt_db.config.container` (as well as `.spt_db.config.local`) from the build directory to the test directory.

More broadly, the two `db/.spt_db.config.[container | local]` files get a lot of usage across many build and test scripts, and their new location in the build directory is likely the cause of multiple test failures. They have to be located in the same directory as the db Docker build so that the latter will work properly, so references will need to be updated to that location if that's where we want to keep them. If not, we'll end up with another case of the `apiserver app COPY` issue.

CarlinLiao:
After further analysis, all remaining failed builds or tests are due to one of two previously enumerated reasons:
* `apiserver` Makefile can't find `app/` folder because they no longer share a directory (former is in `build/apiserver/` latter is in source AKA `spatialprofilingtoolbox/apiserver/`)
* Tests that involve spinning up a database can't find `.spt_db.config.container` because it got moved to `build/db/` instead of staying in the source directory `spatialprofilingtoolbox/db/`.

The former we need to figure out how to inject external directories into a Makefile's context, either by messing with the Makefile to include a copy command or injecting additional context into `docker run`. The latter could be solved by something as simple as copying the config files everywhere they're needed, or by the same solution as for the former problem.

CarlinLiao:
Simply having two copies of `.spt_db.config.container`, one in `build/db` and `test/db`, fixed the errors. I think this should be acceptable since it's just a short file containing a bunch of database credentials. (I doubt we want to do this for `apiserver/app/`, which is the only remaining issue.)

CarlinLiao:
The rule to build the completions script runs in front of every other rule. Is that necessary?

CarlinLiao:
A lot of rules require the absolute location instead of the relative location to work (e.g., `docker run`, submodule Makefiles). I think the only reason we're using relative paths at all is because that's the only way using `sed` to swap one directory for another doesn't error, since we can't handle `\` characters _inside_ variables being passed to `sed`.

CarlinLiao:
Re: the app folder issue, [this is a StackOverflow post with a lot of clarification about how the build context works.](https://stackoverflow.com/questions/27068596/how-to-include-files-outside-of-dockers-build-context) I think our best bet is to just copy `app/` to `build/` as part of the build-docker-images rule, and then delete it after the docker image's built.

jimmymathews:
I will figure out a good way to get the `/app` contents available to the Docker image, probably from the wheel directly.

jimmymathews:
I think this is fixed up by adding this argument to the `uvicorn` call:

```
... --app-dir /usr/local/lib/python3.9/site-packages/spatialprofilingtoolbox/apiserver/ ...
```
Adding this now.

jimmymathews:
Tests pass for me.

jimmymathews:
Looks good, I approve.
A last note: Shall we place `build_scripts` in `build` just to reduce the number of subdirectories at top level?

jimmymathews:
I actually might be getting an error in the last test (workflow), after deleting a bunch of old artifacts (the new make clean doesn't delete them). Let me see...

jimmymathews:
Tests pass for me with the new commits.

Issue #108, opened 2023-01-10
Separate completions script building


jimmymathews:
This package uses a CLI command pattern that requires, at packaging time, a certain minimal/skeletal knowledge of the rest of the Python package. Modules that provide commands provide those commands' names using a Python function, and at packaging-time these are called in order to create a bash completions script to be distributed as part of the actual package.

This led to a poor workaround: importing the Python package without actually packaging it, just in order to run those "get command names" functions and to chart the module structure that should be mirrored in the CLI.

We have a few options to avoid this workaround.

One option (that well-preserves the current functionality):

1. Implement a declaration pattern that provides everything that the CLI completion build needs in a single static data artifact.
2. Rewrite the "get commands" and "get modules" stuff to rely solely on this static artifact.
3. Move all the CLI completion *build* stuff out of the primary source tree.
4. Add a test that checks that the static data artifact is consistent with what the package actually provides.

This would be a kind of a machine-readable-API pattern.

A simpler option is:

1. Omit the build-time/packaging-time step of creating the bash completions script.
2. Do the creation of the bash completions script at install time, in `spt-enable-completions`.

The downside to the simpler option is that it makes the package less inspectable. You won't know what bash script will get installed to your `~/bin` or whatever until you actually run it, even though there is no customization or anything that necessitates this. I think this is nevertheless acceptable.

So for this issue, do the simpler option above.

jimmymathews:
Completed with #109 .

PR #109, opened 2023-01-10
Deprecate completions script build step

jimmymathews:
This addresses #108 .
Implementation notes:
- All of the appearances of completions script references in the Makefile are gone.
- The Makefile in `entry_point` is deprecated.
- The completions script is built at install time in `spt-enable-completions`.
- The `spt-completions.sh` package data file is deprecated (and `spt-completions.sh.jinja` has been added as a package data file).
- A test was added, out of the source tree (so it is not currently called during `make test`, due to outstanding #103 ). The test just checks that the completions script installation matches a saved instance of this script. This will need to be updated when new modules or commands are added.


CarlinLiao:
Three notes before merging:

- I've tried to run the test on my machine but I don't know how to do so in the intended way. I don't see a corresponding Docker image for it, and running the test directly it complains `spt-enable-completions not found`.
- Will this require any updates to `docs/maintenance.md`?
- IMO the source files `spatialprofilingtoolbox/entry_point/spt*` should be renamed to something more clear about their purpose, e.g., `cli`. `spt.py` implies that it's the lynchpin source file of the entire package, when that's not the case.

jimmymathews:
(I had planned for the proper test running for this test to wait for #103 ).

CarlinLiao:
Got it, the test passes on my machine now that I know the requirement. Functionally this is good to merge, although I would suggest a rename of all those files to `cli.py` and similar. (Reminds me of the old Microsoft Word assistant, but alas.)

Issue #107, opened 2023-01-10
Workflow/workflows terminology


jimmymathews:
At one time this repository consisted primarily of the code currently located at `spatialprofilingtoolbox/workflow/workflows`.

Each subdirectory there had all of the components needed to run a Nextflow script that would import datasets, run potentially parallel jobs over them, and aggregate results.

(Many of these "workflows", as I termed them, borrowing some of Nextflow's terminology, have still not been brought up to speed with respect to the updated/cleaned up data model, due to a few ouststanding blocking issues. Nevertheless they will eventually be updated and this will be the place where "backend" computation is done, and new algorithms added. E.g. the cggnn stuff, or external-library spatial metrics.)

As the services located in this repository multiplied, these workflows became just one component. At the time of deciding on names for these components and services, I chose the name "workflow" for the module that mainly contains the workflows.

In long paths this seems redundant, but it is not actually redundant, because it identifies each subdirectory of `workflows` as one particular workflow. We could, for the sake of convenience/readibility alone, excise the `workflows` directory and make its subdirectories into direct subdirectories of `workflow`.

The downside is that this makes each workflow module sibling to `templates`, `source_file_adi_parsing`, and `scripts`, which are not workflows. That is confusing, but not fatal, so for this issue excise `workflows`.

jimmymathews:
There is a conflict between the values of short monikers (codified somewhere in Python's suggested package and module/file name conventions) and descriptive names (not as formally valued in Python documentation and conventions). `spatialprofilingtoolbox/workflow/workflows/...` is one manifestation of this conflict.

jimmymathews:
The `source_file_adi_parsing` could eventually move out to its own module, perhaps "data import" or something. This is a point in favor of the notion that making workflows' directories sibling to `source_file_adi_parsing` is OK.

PR #110, opened 2023-01-11
Workflow/workflows terminology simplification

jimmymathews:
Addresses #107 .

Functionality should be unchanged. Mostly a superficial name change, just in a lot of source files.


jimmymathews:
(The list of commits includes some seemingly unrelated commits from a separate closed issue/PR... this is due to my habit of merging in pending work. I'll try to avoid this in the future.)

Issue #105, opened 2023-01-09
Fix linter complaints


CarlinLiao:
Per #84, here are the linter complaints that have yet to be addressed:

* pylint
   * prefer f-strings to `%` string formatting
   * use lazy `%` formatting in logging functions
   * `open` call doesn't specify encoding
   * constant name doesn't conform to UPPER CASE spelling style
   * redefine [variable name] from outer scope
   * complaint about the way an if statement is ordered
   * `!= None` -> `is not None`
   * consider iterating over a dict directly instead of calling keys()
   * unused variable
      * (in tuple returns where only some components are used)
      * (in a try-except)
      * (in a test)
      * (if it's the only line in a function)
   * consider using enumerate instead of range and len
   * consider iterating over a dict using items instead of over key lookups
   * variable name doesn't conform to snake_case
   * ~line too long in the case of a string that doesn't have an obvious breakpoint (e.g., URLs and regex, whereas SQL statements are easily breakable)~ let's leave them as-is
   * unused argument
   * anomalous backslash (on regex expressions)
   * hyphenated .py filenames instead of snake_case
   * imports not grouped (in the case of try/except imports not in the header of a file)
   * no exception type specified
   * dangerous default value ([], {}, etc.)
   * subprocess.run call doesn't explicitly set "check"
   * implictly defined variables in a class and for loop (e.g., check for one variable assures that another variable isn't unassigned)
   * super init not called
   * overriden child class function has more parameters than original
   * `assert condition == False` should be `is False` or `assert not` depending on if you're testing for exactly false or falsiness
   * unnecessary pass in abstract methods
   * using possibly unidentified loop variable
* pylance
   * ~function already defined (`get_phenotype_criteria` is defined thrice but is attached to a different flask endpoint each time)~
   * define new function instead of assigning a lambda a variable name
   * defining class attribute outside of init
   * [class instance] has no [attribute] (when creating the new property by assigning to it)
   * argument has incompatible type (when calling a builtin function that is correctly type hinted with a user-defined variable that isn't)
   * incompatible type when assigning to a dict
* spellcheck: [many variable names] is not a word (although interventional is definitely not a word)

# Acceptance

All issues on this list are fixed, or have an explanation added as to why we've chosen not to fix it.

CarlinLiao:
Should I rename dashed python script names to underscores? @jimmymathews

jimmymathews:
Yeah these can be converted to dashed CLI command names by pyproject.toml spec I think.

[...]

Issue #100, opened 2022-12-14
Add type hints and docstrings


CarlinLiao:
Per #84 and #86, add docstrings and type hints to all Python files to comply with pep8 standards and make the codebase more readable.

# Acceptance

* All Python modules, classes, and functions have at least one-line docstrings.
* All Python methods have their function calls fully type-hinted and are implemented free of misaligned type calls according to mypy linting.

CarlinLiao:
I shouldn't be surprised but there's also a convention for Python docstrings, [PEP 257] and a linter to support those conventions called [pydocstyle.](http://www.pydocstyle.org/en/stable/) We should consider adding this as a development requirement.

jimmymathews:
I added a disable flag to `.pylintrc` for function-level docstrings. I don't think it is helpful to add these uniformly across the whole codebase, it is likely to be a formality in the majority of cases (with uninformative 1-liner docstrings).

As part of #115 class and module level docstrings are all present. Could close with #115 merge.

CarlinLiao:
[New tech for better documentation.](https://www.jwz.org/blog/2023/02/code-with-swearing-is-better-code/)

PR #115, opened 2023-01-14
Fix (most) linter complaints

CarlinLiao:
Pursuant to #105.


CarlinLiao:
I've finished looking at every Python file in the source folder.

Complaints not yet silenced that aren't covered by #100:
* "Unused argument 'kwargs'" in workflow class inits like Initializer and ComputationalDesign. I removed them initially but things started erroring because when Nextflow builds the Docker containers, `spt workflow run ...` passes all input arguments to every instantiation, and if the class init can't handle input args it doesn't use, it'll error.
* "Dangerous default value [] as argument" because I'm not sure how to fix this and retain functionality.
* "Formatting a regular string which could be a f-string" for nested replacements like `('{0:0%sb}' % str(number_channels)).format(binary)`
* "Line too long" for really long imports and logging formatted strings that can't be reasonably broken up into short enough lines.
* "No exception type(s) specified" in `workflow/source_file_adi_parsing/sample_stratification.py` because the Exception is used to catch if ANY valuation method fails by erroring, so it would have to be replaced by an exhaustive list of all possible exception types we'd consider unreadable as opposed to failing for another reason.
* "Variable name ... doesn't conform to snake_case naming style" for data science-y variable names (e.g., `n` and `df`)
* A couple more I think?

There were also one or two functionality issues I identified but I lost my issue draft so we'll just have to rediscover them.

jimmymathews:
- For the unused kwargs, let us keep this as is for now. As we saw, this is a fairly deliberate use of the current Python functionality for processing keyword arguments without explicitly naming them. If Python will deprecate this, it is a loss for the possiblity of using the "mixin" pattern. Just something to keep an eye on for the future.
- My usage of [] as a default value is just poor man's type hints, or possibly unnecessary use of the mutable variant (maybe could use empty tuple instead). These could be fixed by using type hints and/or empty tuples I think.
- This particular usage of general-purpose exception is weird... one of the arguments to that function is a function, and the point is to see if this function cannot be evaluated on the other argument, due to an exception. For definiteness, we can fix this up by coming up with a few non-examples, running at the Python CLI, and noting the exceptions that occur and catching these.
- Like the linter, I also don't like the data-sciencey shortnames as a general rule (preferring super-verbose phrases), but I think the ones that remain are probably there for a good reason (e.g. `df` is the conventional name in pandas-land).

As for the unbound `fov_index` and `table`, this is indeed a mistake! The fov indices are meant to be deprecated, but in some places they persistent formally (I think when there is just 1 fov index to loop over, so no error should occur). Let's make this a separate issue or bug and not allow it to block this PR.

jimmymathews:
Finally started using VSCode. Started going file by file addressing these issues.

spatialprofilingtoolbox/db/scripts/retrieve_feature_matrices.py
```diff
@@ -30,12 +34,13 @@
...
     if args.database_config_file:
         database_config_file = abspath(expanduser(args.database_config_file))
-    if not exists(database_config_file):
-        raise FileNotFoundError(
-            'Need to supply valid database config filename: %s', database_config_file
-        )
+        if not exists(database_config_file):
+            raise FileNotFoundError(
+                f'Need to supply valid database config filename: {database_config_file}'
+            )
```
CarlinLiao, line 40: Shouldn't everything below this line also be moved into the if statement? They require the `database_config_file` to exist or it'll error.

CarlinLiao:
[Regarding lists as dangerous default values, the docs say you should have the default value be None and test for it in the body of the function.](https://stackoverflow.com/questions/366422/what-is-the-pythonic-way-to-avoid-default-parameters-that-are-empty-lists)

CarlinLiao:
While linting the tests I noticed a few things:
1. you might know this already, but you can call functions like `tuple`, `set`, and `all` directly on a ~list~ comprehension (generator) statement without wrapping a list around it first.
2. What's up with `test/countsserver/module_tests/benchmark-partial-counts'? It's a python script but the extension isn't in the filename.
3. See #126

spatialprofilingtoolbox/db/scripts/retrieve_feature_matrices.py
jimmymathews: I think the behavior I had in mind was that the first block is argument parsing and the rest assumes parsed args. That is, instead of indenting everything below to fit it inside the if statement, I'd rather add an `else` clause that handles the (similar but different) error that `args.database_config_file` isn't even defined.

jimmymathews:
Yeah I'm aware that generators are the right thing, I've just had bad experiences wasting hours debugging them, involving type errors, tuples vs lists or something, maybe due to the parentheses syntax `(` rather than `[`. Would happily convert these where possible! An advantage is shorter loop time (for boolean expressions like `all` and `any`), and a disadvantage is lack of standard "storage and retrieval" API (`g[3]` is impossible for generator `g`).

`benchmark-partial-counts` was a script I wrote to test the fast counts service to see if it is was actually better than the predecessor functionality. It is. I did not want to destroy the script just because it isn't needed in the library. It isn't needed even as a normal development test, but it could nevertheless eventually be useful (e.g. when publishing about this work).

jimmymathews:
I am adding the following to my `autopep8` formatter arguments in vscode: `--max-line-length=100`.
This is consistent with the default setting for `pylint`.
The default (because it is part of PEP8, I guess) is maximum line length 80, but this is too short. Having a large number of lines weirdly broken is too hard to read in my opinion. When I encounter the weirdly broken stuff I'm going to change it back unless it's over 100...

jimmymathews:
I'm having trouble getting vscode pylint or pylance to actually show the "refactor" level or "warning" level stuff. I can only find these by running pylint manually.
I'm using "Go To Next Problem /..." to iterate through them in case they aren't visually obvious, and with this method it appears there are no problems when in fact there are many problems.

jimmymathews:
There seem to be multiple places in vscode settings related to problem severity.
Eventually I found a mapping from pylint levels to vscode display severity (searching "severity" in settings), and messing with these worked.

jimmymathews:
I did a big sweep over the whole codebase, informed by the CLI whole-package `pylint` output. There were perhaps a few hundred minor items, mostly R-level ("refactor"). `pylint` now reports:

| type       |number |
|------------|-------|
| convention |1      |
| refactor   |19     |
| warning    |0      |
| error      |0      |

| message id             |occurrences |
|------------------------|------------|
| too-many-arguments     |7           |
| too-many-locals        |6           |
| too-many-branches      |3           |
| too-many-statements    |2           |
| too-few-public-methods |1           |
| missing-final-newline  |1           |

This included:
- Fixing the bug where docker image data preloaded failure would not be noticed.
- Deprecating the front proximity workflow.
- Fixing lots of duplicate/similar code blocks.
- I think everything mentioned in this thread and the related issue description.

The remaining problems are too many statements / branches / arguments, etc. These are important but require much more work to fix, since this entails breaking up large functions, coining new aggregated data types, etc. For now let us leave these as they are.

Test pass so I think this PR is ready for review.

jimmymathews:
pylint ranks the code at 9.96/10, woo!

PR #131, opened 2023-02-11
update cggnn to new call signature

CarlinLiao:
To match [`cg-gnn#3`](https://github.com/CarlinLiao/cg-gnn/issues/3), I've updated the code that runs when `spt cggnn run` is called to match the new call signature of `cg-gnn`'s `run_pipeline` function, as well as its associated test.

### DIFF
--- a/spatialprofilingtoolbox/cggnn/scripts/run.py
+++ b/spatialprofilingtoolbox/cggnn/scripts/run.py
@@ -12,26 +12,14 @@ except ModuleNotFoundError as e:
 def parse_arguments():
     "Process command line arguments."
     parser = ArgumentParser(
-        prog = 'spt cggnn run',
-        description = 'Create cell graphs from SPT tables, train a graph neural network on them, '\
-            'and save resultant model, metrics, and visualizations (if requested) to file.'
+        prog='spt cggnn run',
+        description='Create cell graphs from SPT tables, train a graph neural network on them, '
+        'and save resultant model, metrics, and visualizations (if requested) to file.'
     )
     parser.add_argument(
-        '--measurement_study',
+        '--study',
         type=str,
-        help='Name of the measurement study table in SPT.',
-        required=True
-    )
-    parser.add_argument(
-        '--analysis_study',
-        type=str,
-        help='Name of the analysis study table in SPT.',
-        required=True
-    )
-    parser.add_argument(
-        '--specimen_study',
-        type=str,
-        help='Name of the specimen study table in SPT.',
+        help='Name of the study to query data for in SPT.',
         required=True
     )
     parser.add_argument(
@@ -138,9 +126,7 @@ def parse_arguments():
 
 if __name__ == "__main__":
     args = parse_arguments()
-    run_pipeline(args.measurement_study,
-                 args.analysis_study,
-                 args.specimen_study,
+    run_pipeline(args.study,
                  args.host,
                  args.dbname,
                  args.user,
--- a/test/cggnn/unit_tests/test_run.sh
+++ b/test/cggnn/unit_tests/test_run.sh
@@ -1,7 +1,5 @@
 spt cggnn run \
-    --measurement_study "Melanoma intralesional IL2 (Hollmann lab) - measurement" \
-    --analysis_study "Melanoma intralesional IL2 (Hollmann lab) - data analysis" \
-    --specimen_study "Melanoma intralesional IL2 (Hollmann lab) - specimen collection" \
+    --study "Melanoma intralesional IL2" \
     --host spt-db-testing \
     --dbname postgres \
     --user postgres \


jimmymathews:
I am getting this test error:
```
  File "/opt/conda/lib/python3.9/site-packages/spatialprofilingtoolbox/cggnn/scripts/run.py", line 129, in <module>
    run_pipeline(args.study,
  File "/opt/conda/lib/python3.9/site-packages/cggnn/run_all.py", line 36, in run_pipeline
    graphs = generate_graphs(df_cell, df_label, validation_data_percent,
  File "/opt/conda/lib/python3.9/site-packages/cggnn/generate_graph_from_spt.py", line 283, in generate_graphs
    graphs_by_label_and_specimen, graph_names = _create_graphs_from_spt_file(
  File "/opt/conda/lib/python3.9/site-packages/cggnn/generate_graph_from_spt.py", line 80, in _create_graphs_from_spt_file
    graph_instance = _create_graph(
  File "/opt/conda/lib/python3.9/site-packages/cggnn/generate_graph_from_spt.py", line 127, in _create_graph
    adj = kneighbors_graph(
  File "/opt/conda/lib/python3.9/site-packages/sklearn/neighbors/_graph.py", line 128, in kneighbors_graph
    return X.kneighbors_graph(X=query, n_neighbors=n_neighbors, mode=mode)
  File "/opt/conda/lib/python3.9/site-packages/sklearn/neighbors/_base.py", line 988, in kneighbors_graph
    A_data, A_ind = self.kneighbors(X, n_neighbors, return_distance=True)
  File "/opt/conda/lib/python3.9/site-packages/sklearn/neighbors/_base.py", line 810, in kneighbors
    raise ValueError(
ValueError: Expected n_neighbors <= n_samples,  but n_samples = 2, n_neighbors = 6
```
Not sure if this really makes sense. Shouldn't the call to `kneighbors(X, n_neighbors, return_distance=True)` see something like 100 cells as some dimension of `X`? Not sure what it considers to be samples, but graph-building should be per-slide.

CarlinLiao:
The main issue seems to be `n_samples`. If I recall correctly, it's the number of cells in the ROI, and for some reason it's only seeing two cells in this ROI, not... many (I think typically it would be hundreds or thousands). The true problem is likely upstream in ROI creation. If I were you I'd put a breakpoint here using the VS Code debugger and investigate the ROI data structure.

CarlinLiao:
Granted, you're running the test in a remote environment (presumably, since this prefers a discrete Nvidia GPU), using a shell command, in a Docker container, that's running a pre-written shell file, so it's harder to do line-by-line debugging...

jimmymathews:
You are right, the upstream problem is creation of an ROI with 2 points:
```
Specimen: lesion 0_1
  Bounding box 0 number points: 70

Specimen: lesion 0_2
  Bounding box 0 number points: 64

Specimen: lesion 0_3
  Bounding box 0 number points: 84
  Bounding box 1 number points: 50

Specimen: lesion 6_1
  Bounding box 0 number points: 46
  Bounding box 1 number points: 37
  Bounding box 2 number points: 29

Specimen: lesion 6_2
  Bounding box 0 number points: 78
  Bounding box 1 number points: 44

Specimen: lesion 6_3
  Bounding box 0 number points: 37
  Bounding box 1 number points: 12
  Bounding box 2 number points: 2
  <this is where it crashes>
```
It seems that this should be fixed by discarding ROIs that are too small to make graphs.

jimmymathews:
The ROIs do seem to be not quite covering properly (this test dataset has 100 cells per slide).
I think the culprit may be this [line](https://github.com/CarlinLiao/cg-gnn/blob/dc207dc558ff6d714156e88a3d749773a114a4d0/cggnn/generate_graph_from_spt.py#L37):
```
slide_size = df_specimen[['center_x', 'center_y']].max() + 100
```
`slide_size` is then used to decide how many ROIs to look for in the next round after picking out each ROI. I think this could be rounding to 0 prematurely.
That line should probably say something like max minus min.

CarlinLiao:
Thanks for diagnosing the root cause, it's very helpful.

`slide_size` is the max x and y values of the slide in pixels, plus 100 pixels to allow an ROI to go past the edge of the slide. It indeed assumes that slides start at (0,0) and go to the max x and y values. It doesn't change as ROIs are found.

I think the problem is ultimately conceptual, rather than in the implementation. My algorithm assumes that, so long as there are enough cells left that another ROI should be created, it assumes that the next ROI it makes is going to be of decent quality. But the ROI will be bad if the remaining uncaptured cells are distributed so sparsely that the next box captures only a few cells. I doubt this happens much in practice since real pathology slides are so dense, but it pops up in this sparse test example.

My remedy would be to add an additional break condition to the while loop such that, if the next ROI created would cause a crash, i.e., if it has fewer cells in it than there are neighbors to find, it discards the ROI, exits the while loop, and moves on to the next slide.

CarlinLiao:
I've updated the `cg-gnn` PyPI package to include this condition and fix some faulty ROI creation logic. I think it should work now, but I wasn't able to test it on my local machine because it kept crashing due to what I'm guessing is insufficient memory. (I never had any issues with this on Saad's machine.) Updated to `cg-gnn>="0.1.5"` and let me know if it fixes the issue.

jimmymathews:
The training goes through and outputs are generated. Took about 50 seconds on my machine.

jimmymathews:
I think it is ready to merge? All tests pass.

Issue #125, opened 2023-01-26
Revive support for continuous intensities


jimmymathews:
We need our datasets to be able to provide "... Intensity" columns and not just "... Positive" (binary/dichotomized) columns.
The data model supports these, but the current "HALO import" workflow still does not attempt to populate them.

For this issue, choose an expected format for CSV/TSV input with intensity values, and get the "HALO import" workflow to parse them (and not just the binary classification values 0 or 1), and store these in the `quantity` field of the `expression_quantification` records.

(*This functionality was once part of SPT but broke with various updates to input data formats. It needs to be repaired in order to make a non-trivial point cloud of the cell set.*)

PR #145, opened 2023-04-10
add intensity parsing to the workflow

danjafish:
solves #125


jimmymathews:
I refactored a little bit to make the column name stuff (underscores vs spaces, etc.) handled in the "dataset design" class, separate from the parsing in `cell_manifests.py`.

I also added 2 tests, one that checks that a few specific intensity values are really in the database after import, and that they match the values in source files, and another test that checks whether datasets are loaded the same way whether or not intensity values are present. The latter test is represented by a new "preloaded image" with a third test dataset, a variant of dataset 1 ("small") with the intensity values removed.

All tests pass for me, @danjafish please check `make test` and verify that these work for you as well.

Note that for the first time you're doing this, you should now either:
- run `make force-rebuild-data-loaded-images`
- run `make clean-docker-images`
before doing `make test`.  This ensures that the data-loaded images are rebuilt using the new import functionality.

danjafish:
@jimmymathews tests worked for me

Issue #143, opened 2023-04-04
On-demand spatial metrics calculation


jimmymathews:
The phenotype proximity metrics calculation is proving to be "heavy" in the context of many datasets. I think I observed that unpredictably (probably due to a few samples with very densely-packed cells) some samples will take >10 hours to process.

This is for a few tens of thousands of phenotype pairs.

For this issue, add an API endpoint that computes this metric for a single pair if not already computed. Hopefully this will be very fast, and can be effectively run inside a timeout wrapper just in case. (Usage of this endpoint will then replace the current frontend's query for all computed values).

jimmymathews:
Being worked on in `issue143`.

jimmymathews:
This turns out to require a lot of RAM resources for the API server container.
Possibly this is due to an unoptimized use of RAM during the pulling of expression data step. E.g. the Mel dataset crashed even at 14G limit, seemingly only about 80% done with the pull. I think the ongoing RAM usage is less.

jimmymathews:
I will plan to alter [get_sparse_entries](https://github.com/nadeemlab/SPT/blob/2936bd24b07d7529e9a79dd0146511c008af13da/spatialprofilingtoolbox/workflow/common/sparse_matrix_puller.py#L85) so that it makes pulls per-specimen only (even in the whole-study case), and pickles them to a local storage like `/tmp/file.pickle`, and somehow alter `parse_data_arrays_by_specimen` so that it doesn't require the full list of sparse entries as a single object. This will hopefully have the side effect of cleaning up the tracking of the specimen identifier during `parse_data_arrays_by_specimen`.

jimmymathews:
Alternative to the above: Perhaps it is better to augment `countsserver`, which already relies on the expression matrices pre-collated, with a few more data files with shape/location data, and add the on-demand functions to this service. This service could even be scaled up, I think, fairly easily.
This may be a low-resistance development path, since `countsserver` already exists and the caching functionality already exists, taking some initialization burden off the running containers.

jimmymathews:
That last option was implemented.

A new API endoint `/request-phenotype-proximity-computation` is available, which sends a request to the `countsserver` (now not so appropriately named) for this calculation.

jimmymathews:
Reopening this to add a new aspect: A minimal task queue that each request for proximity metrics pushes to, so that the requester can poll for intermediate results rather than waiting a super long time for the response.

jimmymathews:
Plan:
[...]

PR #146, opened 2023-04-11
Add on-demand proximity calculation

jimmymathews:
Addresses #143 .
The countsserver service (eventually should be renamed to reflect new functionality) now also does an on-demand one-case computation of the proximity metrics across the specimens in a given study. Exposed via the API server.

All tests pass, and a few were added. Importantly, the test for exact values of the proximity metric still passes. The exact values consistently match the values obtained in:
- an older prior implementation of the proximity workflow, taking CSV source files
- a subsequent implementation of the proximity workflow meant to take the DB as its input
- an on-demand version of the proximity calculation implemented initially in this PR's history, which loads all location data on first access
- the on-demand version of the proximity calculation (proposed in this PR) which, like the countsserver, starts from a pre-created cache to save initialization time


jimmymathews:
New docker images were also released, with version bump.

Issue #119, opened 2023-01-19
New workflow, precomputed static cell set visualization


jimmymathews:
For this issue,

- Add a new workflow that computes UMAP visualization of the entire cell set for a given study, along the phenotypes that are part of the study.
- Add a new table ("out of band", i.e. not part of the main data model) with a base64 encoding of SVG or similar vector graphics representations of: one UMAP for each channel with an overlay of the intensity channel (say, in green to gray).
- Then add an API endpoint that retrieves all the UMAP base64 strings by study.

Note that, to enable the above, first the data import workflow needs to be updated to retrieve the intensity values for each channel and cell (See #125, which blocks this).

(This work is arguably blocked by #83, since #83 should provide a clean template for new workflows.)

danjafish:
@jimmymathews do you expect me to start with the issue #125 or work on this one assuming some dummy values?

danjafish:
@jimmymathews That's an example of UMAP clustering for sample dataset. Here are 7 samples and separate cluster for each sample. But I assume that's not always a case. Can you clarify cluster coloring? What should we use as clusters? 
![image](https://user-images.githubusercontent.com/34601963/231773158-3c4eb91f-54e3-4f99-8ef0-4bab81e12aa5.png)

danjafish:
@jimmymathews I've created a new branch [issue119](https://github.com/nadeemlab/SPT/tree/issue119). The changes are quite significant, so I suggest reviewing them before I submit a PR. Here are a few notes:

The changes have not been tested on large datasets yet.
I've added an API endpoint based on your previous work as an example, but it hasn't been tested. This is another issue we should address.
There are several assumptions made about the workflow structure, so please feel free to provide feedback if you think something needs to be changed.
We should consider adding tests, as there aren't any at the moment.

jimmymathews:
@danjafish : I created `issue119review` branched off of `issue119`. I refactored a bit and added a test that creates an HTML page out of all the images, and compares it with such a page committed to version control.

Please have a look and merge it into `issue119` if you're satisfied, that is:

```sh
git checkout issue119
git merge issue119review
```

Then when you're ready, you can make a PR for `issue119`.

jimmymathews:
(Just realized I didn't look at the API endpoint yet.)

jimmymathews:
[...]

PR #147, opened 2023-05-02
UMAP visualization

jimmymathews:
Addresses #119 .


jimmymathews:
Tests pass, and the resolution/quality looks acceptable now, for example:
image

Issue #151, opened 2023-05-08
Development image optimization


jimmymathews:
The development process includes creating a Docker image that has the whole basket of dependencies pre-installed, everything from sql drivers to python packages like scipy. This image is then used for package building and testing.

Currently the development image make target depends on the spatialprofilingtoolbox source, but only because it creates the package as a side effect.

Docker's build system is pretty smart and figures out most of the time that few layers need to change. An exception is when a version bump changes `pyproject.toml`, because this invalidates the dependency package layers which are named in `pyproject.toml`.

Fix this by creating a separate python package make target, and factor `pyproject.toml` into a versionless source (pre-requisite for the development image) and a make-generated final form. This way the development image is only rebuilt if the dependency layers actually need to be rebuilt/reinstalled.

jimmymathews:
Done in branch `issue151`.

PR #152, opened 2023-05-08
Optimize development image rebuild

jimmymathews:
Addresses #151 .


jimmymathews:
All tests pass. I also checked that source code changes result in only about 10 seconds of time spent rebuilding layers, on repeat of `make test`.

Issue #162, opened 2023-05-24
API docs


jimmymathews:
The API server uses FastAPI, which automatically creates Swagger/OpenAPI documentation.

- Learn FastAPI's system for adding doc information.
- Add in as much documentation as possible for all the endpoints, plus higher/section-level explanations.
- Refactor and rename as necessary, adding additional convenience endpoints, etc.

Eventually we want to support the case of non-interactive scripts meant to be run locally on a local SPT instance, for reproducing analyses, which should mainly get their results via the API server (rather than internal library functions). For this to be not just reproducible but also explainable and interpretable, *what* is performed needs to be very clear.

jimmymathews:
After reviewing the main source file, it's clear that the API handlers need refactoring as well. Much of this is moving accessor functionality into the db module, and could perhaps involve a library of sql queries.

jimmymathews:
I am adding refactoring of the database accessing to the scope of this issue. The main thing is to reduce the rampant duplication between `DBAccessor` and `DatabaseConnectionMaker`. Historically this was caused by the apiserver requiring a different way of providing db credentials, but with a few obvious changes to the design these two different db access providers can be merged.

jimmymathews:
The db access revision is complete, in branch `issue162`.

PR #174, opened 2023-07-12
API refactor and docs

jimmymathews:
Addresses #162.

This PR improves the API considerably, with the end result that the API becomes well-documented and more secure.
- Gathers the various database access wrappers (which handle credentials automatically), replacing them with one accessor. Includes a short syntax for immediate availability of a cursor.
- Shortens the main source file with API handlers from 1000 to 200 lines.
- Moves much of the query logic into submodules of the `db` module. The SQL query / cursor / handler logic are now pretty cleanly separated into 3 layers which do not interfere with each other (e.g. no SQL query appears in `app.main.py`.)
- Uses Pydantic types to structure all response objects ([study.py](https://github.com/nadeemlab/SPT/blob/issue162/spatialprofilingtoolbox/db/exchange_data_formats/study.py) and [metrics.py](https://github.com/nadeemlab/SPT/blob/issue162/spatialprofilingtoolbox/db/exchange_data_formats/metrics.py)). This is a lightweight / informal "schema" that serves a different purpose from the database schema (since the latter is largely a long-term storage format with an eye towards scientific accuracy).
- Makes use of FastAPI's basic built-in response validation via the Pydantic type declarations.
- Cleans up the automatically generated Swagger / OpenAPI documentation of the API endpoints and query parameters.
- Introduces strict query parameter validation using FastAPI's concept of "Dependencies".

The changes are deeply API-breaking, so the frontend will have to change all of its functions which reach out to the apiserver.

Since there are so many changes, if you are reviewing please review the changes as thoroughly as you can.

Please include in your review:
1. Check tests.
    ```sh
    make clean
    make test
    # make test VERBOSE=1
    ```
2. The generated documentation, inspectable for example by doing the following:
    - Assuming `make test` or `make development-image` has already been run, install the main package.
        ```sh
        python -m pip install dist/spatialprofilingtoolbox-0.14.0-py3-none-any.whl
        ```
    - Stand up FastAPI.
        ```sh
        uvicorn app.main:app --app-dir spatialprofilingtoolbox/apiserver/ --host 0.0.0.0 --port 8080 --log-level debug
        ```
    - Check a browser at `http://0.0.0.0:8080/docs` .
3. Linting
    ```sh
    python -m pylint spatialprofilingtoolbox/
    ```


spatialprofilingtoolbox/db/fractions_transcriber.py
```diff
@@ -19,76 +19,73 @@ def describe_fractions_feature_derivation_method():
...
+    connection = database_connection_maker.get_connection()
+    cursor = connection.cursor()
+    cursor.execute('''
+    INSERT INTO data_analysis_study(name)
+    VALUES (%s) ;
+    INSERT INTO study_component(primary_study, component_study)
+    VALUES (%s, %s) ;
+    ''', (name, study_name, name))
+    cursor.close()
+    connection.commit()
```
CarlinLiao, line 34: Here you've unwrapped this statement from the `with` statement, whereas in other places you've newly wrapped statements into `with DatabaseConnectionMaker(` clauses. What's the difference between the two cases?

jimmymathews: I think the explanation is: `db.SchemaInfuser` for whatever reason was made to be a subclass of `DatabaseConnectionMaker` a long time ago, so that its methods can  more directly access the db I guess. And the step of computing and uploading the cell-fractions features (including the work of `transcribe_fraction_features` and indirectly the method you referenced) is a wrap up step for the view refreshing performed by `db.SchemaInfuser`. The `SchemaInfuser` just passes along its `self` to provide db access.
In general there seems to be a degree of freedom related to the granularity at which new database connections or cursors are created and used. In a multi-step process, one connection or one for each step?
There is no real strategy here except an aim for consistency. In the new code of this PR, there is a lot of consistency about this; when the cursor is created, which object carries it as temporary state, etc. In the older code the pattern is less clear.

jimmymathews: To elaborate a bit more: `DatabaseConnectionMaker` is only used if the connection is explicitly desired (e.g. to do commits at certain moments). Otherwise I'm preferred `DBCursor` which hides the connection and gives directly one cursor.

spatialprofilingtoolbox/apiserver/app/main.py
```diff
@@ -1,23 +1,36 @@
...
+    ProximityMetricsComputationResult
+from spatialprofilingtoolbox.db.exchange_data_formats.metrics import UMAPChannel
+from spatialprofilingtoolbox.db.querying import query
+from spatialprofilingtoolbox.apiserver.app.validation import ValidChannel
+from spatialprofilingtoolbox.apiserver.app.validation import ValidStudy
+from spatialprofilingtoolbox.apiserver.app.validation import ValidPhenotypeSymbol
+from spatialprofilingtoolbox.apiserver.app.validation import ValidPhenotype1
+from spatialprofilingtoolbox.apiserver.app.validation import ValidPhenotype2
+from spatialprofilingtoolbox.apiserver.app.validation import ValidChannelListPositives
+from spatialprofilingtoolbox.apiserver.app.validation import ValidChannelListNegatives
```
CarlinLiao, line 29: This is a writer's choice thing, but I am curious why you choose to do imports this way instead of
```
from spatialprofilingtoolbox.apiserver.app.validation import ValidChannel, ValidStudy, ...
```

spatialprofilingtoolbox/db/credentials.py
```diff
@@ -0,0 +1,57 @@
...
+
+logger = colorized_logger(__name__)
+
+@define
+class DBCredentials: #pylint: disable=too-few-public-methods
+    """Data structure for database credentials."""
+    endpoint: str
+    database: str
+    user: str
+    password: str
```
CarlinLiao, line 18: Considered using a `NamedTuple` instead of a class?

spatialprofilingtoolbox/db/database_connection.py
```diff
@@ -120,3 +108,62 @@ def get_and_validate_database_config(args):
...
+@define
+class SimpleReadOnlyProvider:
+    """State-holder for basic read-only one-time database data provider classes."""
+    cursor: Psycopg2Cursor
+
+
+class QueryCursor:
+    """
+    Dispatches to any class methods of the special handler class, that require a cursor, with a
+    cursor for a one-time connection.
```
CarlinLiao, line 140: > Dispatches to class methods of the special handler class a cursor for a one-time connection.

spatialprofilingtoolbox/db/study_access.py
```diff
@@ -0,0 +1,228 @@
...
+from spatialprofilingtoolbox.db.exchange_data_formats.study import StudyComponents
+from spatialprofilingtoolbox.db.exchange_data_formats.study import StudyHandle
+from spatialprofilingtoolbox.db.exchange_data_formats.study import StudySummary
+from spatialprofilingtoolbox.db.exchange_data_formats.study import Context
+from spatialprofilingtoolbox.db.exchange_data_formats.study import Products
+from spatialprofilingtoolbox.db.simple_query_patterns import GetSingleResult
+from spatialprofilingtoolbox.db.cohorts import _get_sample_cohorts
+from spatialprofilingtoolbox.db.database_connection import SimpleReadOnlyProvider
+
+class StudyAccess(SimpleReadOnlyProvider):
```
CarlinLiao, line 21: Why is this class separate from the `Query` class? They're both initialized with the same parameter (study) and return simple queries on the study.

CarlinLiao:
Broadly the PR looks fine to merge, although I can't test its actual functionality at this point in time. I think there could be longer form documentation to explain the architectural choices behind the way the API and DB calls are structured or, more importantly really, how a new user would use them and work their way through the submodules.

spatialprofilingtoolbox/apiserver/app/main.py
jimmymathews: Oh yeah, I do this so that it is easier to scan the imported symbols visually, and remove one and place it elsewhere during refactoring. Also the linter complains about long lines.

spatialprofilingtoolbox/db/credentials.py
CarlinLiao: I noticed you used pydantics `BaseModel` elsewhere.

spatialprofilingtoolbox/apiserver/app/main.py
jimmymathews: One way that I have avoided this repetitious appearance in other source files is by aggregating logical groups of symbols into a class. This could even work for symbols of classes, with a little class-provider class. E.g. here could have `ValidTypes` with `ValidTypes.channel()`, `ValidTypes.study()` etc.

spatialprofilingtoolbox/db/credentials.py
jimmymathews: Yeah the use of pydantic is strictly to work well with FastAPI. After some investigation into the options I generally prefer `attrs` for typing support. But since there are only 2 places I'm using `attrs` maybe the usage of `attrs` should be removed?
Yeah `NamedTuple` seems appropriate here.

spatialprofilingtoolbox/db/database_connection.py
jimmymathews: I'll rewrite this! Pretty confusing.

spatialprofilingtoolbox/db/study_access.py
jimmymathews: To directly answer your question: `StudyAccess` is initialized with `cursor`, not `study`, and `QueryHandler` (which is what I think you meant by `Query`) is a delegator not just to `StudyAccess` but to several other classes.

What's going on here is the following...
There are classes `StudyAccess`, `PhenotypeAccess`, `FractionsAccess`, and `UMAPAccess` which are responsible for querying the database and giving back data with a controlled type, each doing their work in their own scope of responsibility. Each one subclasses `SimpleReadOnlyProvider`, for the purpose of having a `cursor` member readily available to each method without requiring laborious repetition of `cursor` as the first argument to every method.
In order to relieve the FastAPI application routes from having to deal with cursors, and with delegation to different scopes of responsibility, there are middlemen `QueryHandler`(stateless, class methods only), and `QueryCursor` (handles context manager for cursor), and the latter is finally exposed. This is how we get nice syntax like:
```py
@app.get("/phenotype-criteria/")
async def get_phenotype_criteria(
    study: ValidStudy,
    phenotype_symbol: ValidPhenotypeSymbol,
) -> PhenotypeCriteria:
    return query().get_phenotype_criteria(study, phenotype_symbol)
```
which would otherwise look like:
```py
@app.get("/phenotype-criteria/")
async def get_phenotype_criteria(
    study: ValidStudy,
    phenotype_symbol: ValidPhenotypeSymbol,
) -> PhenotypeCriteria:
    with DBCursor as cursor():
        PhenotypesAccess(cursor).get_phenotype_criteria(study, phenotype_symbol)
```
To make this work I have `QueryCursor` add methods to itself on-the-fly during initialization, so mypy was unhappy. To make mypy happy I added explicit redundant declarations of the method names:
```py
get_study_components: Callable
retrieve_study_handles: Callable
get_number_cells: Callable
get_study_summary: Callable
get_cell_fractions_summary: Callable
get_phenotype_symbols: Callable
get_phenotype_symbols_all_studies: Callable
get_composite_phenotype_identifiers: Callable
get_phenotype_criteria: Callable
get_channel_names_all_studies: Callable
retrieve_signature_of_phenotype: Callable
get_umaps_low_resolution: Callable
get_umap: Callable
```
I'm open to alternative patterns for achieving the same simplicity of caller syntax.

spatialprofilingtoolbox/apiserver/app/main.py
jimmymathews: I tried to combine them but it doesn't quite work because these symbols need to show up as type hints, and expressions aren't allowed there. I tried a named tuples variant, a class member variant, everything seems to cause a type checking error. So for now we'll just have this big list of `Valid...`.

CarlinLiao: I would write it as
```
from spatialprofilingtoolbox.apiserver.app.validation import \
    (ValidChannel,ValidStudy,ValidPhenotypeSymbol,ValidPhenotype1,ValidPhenotype2,
     ValidChannelListPositives,ValidChannelListNegatives)
```
The slash isn't necessary if you start the tuple on the same line.

spatialprofilingtoolbox/db/credentials.py
jimmymathews: For now I'm keeping `attrs` in order to make use of its functionality of seamlessly providing `__init__` that does the right thing with arguments (for `SimpleReadOnlyProvider` subclasses). Otherwise we would have to add 5 new `__init__` methods with boilerplate.

I'll still change `DBCredentials` to `NamedTuple` though.

jimmymathews: I think that `NamedTuple` still has to be an explicitly defined `class` though, in order to get proper type checking.

spatialprofilingtoolbox/db/database_connection.py
jimmymathews: Done.

spatialprofilingtoolbox/apiserver/app/main.py
jimmymathews: The thing I don't like about this form is that the scanning task is much harder. E.g.,
*Does this list contain `ValidPhenotypeSymbols`?*
You will no doubt have noticed that in several places, I forced in newlines in enumerations of list items, whether in lists, tuples, argument lists, etc., and this is the reason I do that.

Except for this side effect, it is of course better to do it the above way to save lines and omit needless repetition of the module name. A possible compromise the keeps scannability and reduces module repetition:
```py
from spatialprofilingtoolbox.apiserver.app.validation import (
    ValidChannel,
    ValidStudy,
    ValidPhenotypeSymbol,
    ValidPhenotype1,
    ValidPhenotype2,
    ValidChannelListPositives,
    ValidChannelListNegatives,
)
```
(The final newline before closing parenthesis is my go-to style to make each line item have the same format and to announce the scope level delimited by the closing paren. I use this consistently everywhere I think.)

jimmymathews:
Well most of the updates pertain to internal functionality. So longer form documentation about it would be more for library maintainers than for library users. Much of the docs have been deprecated and will require a big refresh anyway at some later time.

jimmymathews:
Tests pass again for me, so I'll merge. I had hoped to add formal examples to the API path operation documentation, plus expanded docstrings for them, but that can be an independent issue not blocking this PR.

Note that on merge of this PR and release of the docker images, API service-dependent stuff will need updating.

PR #178, opened 2023-07-18
Cell importance upload

CarlinLiao:
Sketch implementation of uploading importance scores assuming you have a CSV of them and a DB to load them too, given a config file.

I can't test its functionality myself, so let me know if anything needs fixing.


spatialprofilingtoolbox/cggnn/scripts/upload_importance.py
```diff
@@ -0,0 +1,49 @@
+"""Uploads importance score output from a cg-gnn instance to the local db."""
+
```
jimmymathews, line 2: Elsewhere, this pattern was for grouping imports except that there is no need for a blank line between the module docstring and the first standard library import. In other places where there is a blank line after the module docstring, it is supposed indicate that there are no standard library imports.

CarlinLiao: My personal preferred style is a line break after all docstrings, I think it looks nicer. I stopped doing so in some places after pydocstyle complained about line breaks after _method_ docstrings, but it doesn't complain about adding a line break after module docstrings.

That said, feel free to change if this isn't SPT style.

spatialprofilingtoolbox/db/importance_score_transcriber.py
```diff
@@ -0,0 +1,87 @@
...
+    Parameters:
+        df: DataFrame
+            One column, `importance_score`, indexed by `histological_structure`.
+        target_var: str
+            What variable the GNN was trained on to produce importance_score.
+        database_connection_maker: DatabaseConnectionMaker
+        n_most_important: int
+            Grab this many of the most important cells from each slide.
+    """
+    # Recover the study from the histological_structure
```
jimmymathews, line 31: I am also not a big believer in in-line comments, since they don't have a clear semantics as source code.
Here I would propose a refactor in which each of these 4 one-liner comments acting as "code block headers" are removed, and replaced with a function call to a function wrapping each block. E.g.:
```py
def recover_study_from_histological_structure(connection, histological_structure):
    return read_sql(f"""
        SELECT
            hsi.histological_structure,
            sdmp.study
        FROM histological_structure_identification hsi
            JOIN data_file df
                ON hsi.data_source=df.sha256_hash
            JOIN specimen_data_measurement_process sdmp
                ON df.source_generation_process=sdmp.identifier
        WHERE hsi.histological_structure={histological_structure}
        LIMIT 1
        ;
    """, connection)['study'][0]
```

spatialprofilingtoolbox/db/importance_score_transcriber.py
```diff
@@ -0,0 +1,87 @@
...
+            JOIN specimen_data_measurement_process sdmp
+                ON df.source_generation_process=sdmp.identifier
+        WHERE hsi.histological_structure={df.iloc[0,0]}
+        LIMIT 1
+        ;
+    """, connection)['study'][0]
+
+    study_indicator: str = f'cell importance ({target_var})'
+    if data_analysis_study_exists(database_connection_maker, study, study_indicator):
+        logger.debug('Fractions study already exists for %s.', study)
```
jimmymathews, line 49: Hm maybe this should be a warning. The user probably should be reprimanded slightly for trying to do this multiple times and forgetting that they did so already.

spatialprofilingtoolbox/db/importance_score_transcriber.py
```diff
@@ -0,0 +1,87 @@
...
+        SELECT
+            hsi.histological_structure,
+            sdmp.specimen
+        FROM histological_structure_identification hsi
+            JOIN data_file df
+                ON hsi.data_source=df.sha256_hash
+            JOIN specimen_data_measurement_process sdmp
+                ON df.source_generation_process=sdmp.identifier
+        ;
+    """, connection).set_index('histological_structure').loc[df.index, 'specimen']
```
jimmymathews, line 64: Here we see the real benefit of proper dataframe vectorization, as you proposed.

spatialprofilingtoolbox/db/importance_score_transcriber.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Group by specimen and get the top n_most_important cells
+    df_n = df.groupby('specimen').head(
```
jimmymathews, line 67: How about `df_most_important`?

spatialprofilingtoolbox/db/importance_score_transcriber.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Upload to db
+    das = insert_new_data_analysis_study(
+        database_connection_maker, study, study_indicator)
+    with ADIFeaturesUploader(
+        database_connection_maker,
+        data_analysis_study=das,
+        derivation_and_number_specifiers=(
+            'For a given target variable, a trained GNN identifies how important each cell is to '
+            'the specimen\'s predicted classification.',
```
jimmymathews, line 79: The expected format for this description text is a (possibly very composite) noun phrase which refers primarily to the feature values. It is also allowed to refer and should refer explicitly to the enumerated specifiers. The sentence provided here is instead a factual statement about a different subject, the trained GNN, and leaves open how to understand the feature values. (Are they importance scores? Ranks?)
I propose something like:
```txt
For a given cohort stratification variable (the specifier),
the integer rank of each cell (the subjects of the
feature) with respect to the importance scores derived
from a GNN trained on this variable.
```

spatialprofilingtoolbox/cggnn/scripts/upload_importance.py
jimmymathews: Ah I see, so it is not import-related. The pattern I described above was just something I made up when I was trying to choose a convention. I don't have much of a justification for it.
I'm fine with any choice here, blank line or not after any docstrings. I am happy to conform to your preference here from now on.

spatialprofilingtoolbox/db/fractions_transcriber.py
```diff
@@ -80,20 +45,22 @@ def transcribe_fraction_features(database_connection_maker: DatabaseConnectionMa
...
         ) as feature_uploader:
             values = fraction_features_study['percent_positive'].values
             subjects = fraction_features_study['sample']
             specifiers = fraction_features_study['marker_symbol'].values
             for value, subject, specifier in zip(values, subjects, specifiers):
-                feature_uploader.stage_feature_value((specifier,), subject, value / 100)
+                feature_uploader.stage_feature_value(
+                    (specifier,), subject, value / 100)
 
         perform_tests(das, connection)
```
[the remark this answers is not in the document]
jimmymathews: Woops, I meant this comment to apply to the occurrence of `perform_tests` in the new source file below.

jimmymathews:
If you are OK with some of the cosmetic changes I proposed, I am happy to implement them and save you some time.

I'll also add some tests of the new functionality.

jimmymathews:
I am also ok with mixed styles/conventions, as a little bit of source code authorship archaeology for the archaeologists of the future.

CarlinLiao:
I only see the one comment/change proposed about a blank line after the module docstring.

jimmymathews:
Woops I'm misusing the review feature I think.

spatialprofilingtoolbox/cggnn/scripts/upload_importance.py
```diff
@@ -0,0 +1,49 @@
...
+        description='Upload importance score output from a cg-gnn model to db.'
+    )
+    parser.add_argument(
+        '--filename',
+        type=str,
+        help='',
+        required=True
+    )
+    parser.add_argument(
+        '--target_var',
```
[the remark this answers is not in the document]
CarlinLiao: I guess it's the cohort stratification. It's meant to be whatever target variable the GNN was trained to classify for. I suppose it should be contained in the output artifact, but I didn't design the GNN pipeline to behave like that (yet), and it would make the output no longer a CSV but something more complicated (a JSON or a zipped folder containing text and a CSV)?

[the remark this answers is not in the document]
jimmymathews: Hm well why don't we just add a call to `spt cggnn upload-importance`, with a string variable providing this argument value, to the end of the training? Then the "user" doesn't need to worry about it.

spatialprofilingtoolbox/db/create_data_analysis_study.py
```diff
@@ -0,0 +1,49 @@
...
+    cursor = connection.cursor()
+    cursor.execute('''
+        SELECT das.name
+        FROM data_analysis_study das
+        JOIN study_component sc
+            ON sc.component_study=das.name
+        WHERE sc.primary_study=%s
+    ;
+    ''', (study,))
+    names: List[str] = [row[0] for row in cursor.fetchall()]
```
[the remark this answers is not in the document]
CarlinLiao: In earlier versions of Python, something like `list[str]` would error, so you'd need to get the capitalized version of List that supports nested type hinting, but I guess that's been rectified in the more recent Python versions.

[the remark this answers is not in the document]
CarlinLiao: This is how you wrote the function originally in `frations_transcriber.py` in main branch before I pulled it out to a new `create_data_analysis_study`, so I just left it as is.

spatialprofilingtoolbox/db/create_data_analysis_study.py
```diff
@@ -0,0 +1,49 @@
...
+    if any(re.search(indicator, name) for name in names):
+        return True
+    return False
```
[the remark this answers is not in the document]
CarlinLiao: This is how you wrote the function originally in `frations_transcriber.py` in main branch before I pulled it out to a new `create_data_analysis_study`, so I just left it as is.

spatialprofilingtoolbox/cggnn/scripts/upload_importance.py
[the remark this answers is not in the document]
CarlinLiao: I don't follow. What is upload-importance calling that will provide the stratification variable name?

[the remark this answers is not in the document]
jimmymathews: I'm saying, we can add:
```sh
spt cggnn upload-importance --target-var="$TARGET_VAR"
```
to whatever script does training, and make sure that `TARGET_VAR` environment variable is set correctly by the caller script. I guess I'm suggesting that there does not need to be a separate manually run feature export step like this, it can be automatic.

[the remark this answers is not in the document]
jimmymathews: Just like the `fractions_transcriber` does its stuff right after views refreshing.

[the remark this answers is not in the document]
jimmymathews: So, I guess here in [`run_sql.py`](https://github.com/nadeemlab/SPT/blob/3e8c2a6f2afc05c41044878e4b63481f0c165094/spatialprofilingtoolbox/cggnn/scripts/run_sql.py#L148).
Could skip the CLI step and just directly call `transcribe_importance`.

spatialprofilingtoolbox/db/importance_score_transcriber.py
```diff
@@ -0,0 +1,87 @@
...
+                          target_var: str,
+                          database_connection_maker: DatabaseConnectionMaker,
+                          n_most_important: int = 1000) -> None:
+    """Upload importance score output from a cg-gnn instance to the local db.
+
+    Parameters:
+        df: DataFrame
+            One column, `importance_score`, indexed by `histological_structure`.
+        target_var: str
+            What variable the GNN was trained on to produce importance_score.
```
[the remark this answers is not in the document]
CarlinLiao: I can't think of one. Both locations appreciate the explanation. We could just point this doc to the CLI documentation, but that's not organized for easy reading like this is.

CarlinLiao:
Latest commits should address most notes raised. Still a little confused on how you want to handle ~~`target_var`~~ `cohort_stratifier` functionality, so I just changed the name and left it as is.

jimmymathews:
No problem, tomorrow I'll make an edit and show you what I'm thinking.

spatialprofilingtoolbox/db/create_data_analysis_study.py
```diff
@@ -1,50 +1,61 @@
...
+    specifier: str
+    name: str | None
+
+    def __init__(self, connection, study: str, specifier: str):
+        self.connection = connection
+        self.study = study
+        self.specifier = specifier
+        self._retrieve_existing_study()
+        if self._already_exists():
+            raise ValueError(f'Data analysis study "{self.name}" already exists.')
```
CarlinLiao, line 26: Earlier functionality was to warn if a user tried to create a DAS that already exists. Now it's an error?

spatialprofilingtoolbox/cggnn/scripts/run_sql.py
```diff
@@ -128,6 +134,20 @@ def parse_arguments():
...
 
 
+def retrieve_importances() -> DataFrame:
+    filename = join('out', 'importances.csv')
+    return read_csv(filename, index_col=0)
+
+
+def save_importances(_args):
+    df = retrieve_importances()
+    cohort_stratifier = _args.target_column # Is this right?
```
CarlinLiao, line 144: The target_column isn't what the GNN is trained on, it's what's used to generate ROIs. For example, in one of the melanoma datasets, the target_column indicates whether or not a cell is cancerous, so we create ROIs focused on cancer cells only, not cells overall. The GNN is trained on patient level outcomes, which are acquired via hardcoded SQL query. We would need to change how the `cggnn` pip module acquires labels to make it a CLI argument here.

spatialprofilingtoolbox/cggnn/scripts/run_sql.py
```diff
@@ -128,6 +134,20 @@ def parse_arguments():
     return parser.parse_args()
 
 
+def retrieve_importances() -> DataFrame:
+    filename = join('out', 'importances.csv')
+    return read_csv(filename, index_col=0)
```
CarlinLiao, line 139: Should this feature be available in the DataFrame SPT workflow? That should be the preferred cggnn use option.

spatialprofilingtoolbox/db/create_data_analysis_study.py
```diff
@@ -1,50 +1,61 @@
 """Functions to create, insert, and check for a data analysis study in the database."""
-
+import re
 import datetime
 
-from spatialprofilingtoolbox.db.database_connection import DatabaseConnectionMaker
+from psycopg2.extensions import connection as Psycopg2Connection
```
CarlinLiao, line 5: `connection` is renamed in this module but not in the same way as in `importance_score_transcriber`

spatialprofilingtoolbox/db/fractions_transcriber.py
```diff
@@ -3,8 +3,7 @@
 import pandas as pd
 
 from spatialprofilingtoolbox.db.database_connection import DatabaseConnectionMaker
```
CarlinLiao, line 5: `create_data_analysis_study` and `importance_score_transcriber` have been changed to use psycopg2 connections directly, but not `fractions_transcriber`

spatialprofilingtoolbox/db/importance_score_transcriber.py
```diff
@@ -48,23 +26,45 @@ def transcribe_importance(
         cohort_stratifier: str
             Name of the classification cohort variable the GNN was trained on to produce
                 the importance score.
-        database_connection_maker: DatabaseConnectionMaker
+        connection: psycopg2.connection
```
CarlinLiao, line 29: Reference to module is more clear, but inconsistent with import. Should we adopt this style in the future?

spatialprofilingtoolbox/cggnn/scripts/run_sql.py
CarlinLiao: [Specifically, see this function in cggnn.
](https://github.com/CarlinLiao/cg-gnn/blob/677c343398a2b3ea182f339390e173c5ecdf1d4d/cggnn/spt_to_df.py#L169-L185)
```
def _create_label_df(conn, specimen_study: str) -> Tuple[DataFrame, Dict[int, str]]:
    """Get slide-level results."""
    df = read_sql(f"""
        SELECT 
            slide,
            d.result
        FROM histology_assessment_process hap
            JOIN specimen_collection_process scp
                ON scp.specimen=hap.slide
            JOIN diagnosis d
                ON scp.source=d.subject
        WHERE
            scp.study='{specimen_study}';
    """, conn).set_index('slide')
    label_to_result = {i: res for i, res in enumerate(
        sort(df['result'].unique()))}
    return df.replace({res: i for i, res in label_to_result.items()}), label_to_result
```

jimmymathews:
I have added a test, and so I am able to start properly debugging. There are several (minor) issues with feature uploading, mainly call signature mismatch.

spatialprofilingtoolbox/db/create_data_analysis_study.py
jimmymathews: Hm you're right. Fixing.

spatialprofilingtoolbox/cggnn/scripts/run_sql.py
jimmymathews: I see, it does not make use of the `sample_strata` that we worked out. It uses `diagnosis` directly, which probably isn't right.

In this case, it seems that there is no need for *this* new functionality to be made aware of some specific "target variable the GNN was trained on"; it should be already known as part of the study.

spatialprofilingtoolbox/cggnn/scripts/run_sql.py
jimmymathews: You mean the functionality that pulls out DataFrames from the database? (`FeatureMatrixExtractor`)

I'm not sure, because that functionality is agnostic to files or filenames.

jimmymathews: But yes I copied that filename from `cggnn` source, there is probably a better way like cggnn exposing this filename. In case filename business is what you are drawing attention to.

spatialprofilingtoolbox/db/create_data_analysis_study.py
jimmymathews: Yes, elsewhere in the codebase I've used this form. I'm not opinionated about which one, but we should be consistent as you're suggesting. Which do you prefer? `Connection` or `Psycopg2Connection`?

spatialprofilingtoolbox/cggnn/scripts/run_sql.py
CarlinLiao: I mean that this PR as is does not implement importance uploading if you run the SQL-less workflow.

jimmymathews: So by `DataFrame SPT workflow` you mean `DataFrame (as opposed to SQL) SPT cggnn workfow`?

Yes, in the SQL-less workflow there is no importance score uploading I suppose. And that is expected behavior? Since there is nowhere for these scores to go.

spatialprofilingtoolbox/db/create_data_analysis_study.py
CarlinLiao: `Connection` for brevity, `Psycopg2Connection` to minimize ambiguity.

jimmymathews: Resolved.

spatialprofilingtoolbox/cggnn/scripts/run_sql.py
CarlinLiao: Resolved in live conversation. `run` and not `run_sql` is the preferred workflow.

Issue #176, opened 2023-07-14
Add squidpy features


CarlinLiao:
Per in person discussion, we'll be adding a new set of calculated features that are derived using the squidpy library:
* [`gr.nhood_enrichment`](https://squidpy.readthedocs.io/en/stable/api/squidpy.gr.nhood_enrichment.html#squidpy.gr.nhood_enrichment)
* [`gr.co_occurrence`](https://squidpy.readthedocs.io/en/stable/api/squidpy.gr.co_occurrence.html#squidpy.gr.co_occurrence)
* [`gr.ripley`](https://squidpy.readthedocs.io/en/stable/api/squidpy.gr.ripley.html#squidpy.gr.ripley)
* [`gr.spatial_autocorr`](https://squidpy.readthedocs.io/en/stable/api/squidpy.gr.spatial_autocorr.html#squidpy.gr.spatial_autocorr)

Features that require two clusters (e.g., co_occurrence) will be queried for and displayed separately from other metrics.

Execution process
1. Write standalone function that accepts a DataFrame of cell x and y positions (and cluster identifier in the case of co_occurence) and returns a vector of squidpy metrics
2. Integrate above function into `db` and `schema`
3. Make above available to the web API and add tab to SPT website

CarlinLiao:
I suppose this is a duplicate of / clearer plan for #120.

jimmymathews:
If you are working on this, let's work in branch `issue176`. I may be able to contribute as you are doing this.

CarlinLiao:
I don't see a branch `issue176` in the remote.

jimmymathews:
It doesn't exist yet!

CarlinLiao:
How do you decide whether to wrap ondemand functions in a class or not?

For example, why is `ProximityProvider` a class instead of a suite of standalone functions that take a data object of cell locations and output proximities a la carte? Should our squidpy metrics follow this same pattern and instantiate an instance of the `SquidpyMetrics` class, or could it just exist as something that takes in cell locations and phenotypes and outputs a metric vector?

jimmymathews:
There are two major reasons that I tend to consider as justification for promotion to a class:
1. There are many functions all supporting the functionality of one or a handful of entrypoint functions. Here most functions will be static or class methods. The idea is that future refactorers will be prevented from accidentally breaking up co-dependent functionality, and it is easy to bring all functionality to another context with one import (the class). It's a modularity pattern, separation (and isolation) of concerns.
[...]

Issue #182, opened 2023-07-27
`DatabaseConnectionMaker` backward compatibility with `database_config_file`


CarlinLiao:
Newer SPT modules use the `DatabaseConnectionMaker` to, well, connect to the database, but older modules take in a `database_config_file` string directly to create a connection themselves. This leads to unintended consequences when an older module is used in the newer module, as the `DatabaseConnectionMaker` can't be passed to modules asking for a `database_config_file`.

I ran into this most recently in #176 [here](https://github.com/nadeemlab/SPT/issues/176#issuecomment-1650626006) as the new module I'm writing takes in a `DatabaseConnectionMaker` in its init, but needs to call `FeatureMatrixExtractor` which uses `database_config_file` instead.

As I see it, there are a couple possible fixes for this:
* Rewrite older modules to use the `DatabaseConnectionMaker` pattern
* Write new modules that use older modules to use `database_config_file` only
* Add a property to `DatabaseConnectionMaker` that stores the `database_config_file` used to create the maker

CarlinLiao:
Specifically for `FeatureMatrixExtractor`, refactoring looks nontrivial. It uses `database_config_file` multiple times as it runs through the extraction process, with each step using its own subclass of `DatabaseConnectionMaker` to extract the relevant features. We would need to refactor `SparseMatrixPuller`, etc. to be able to accept an instance of its own parent class, `DatabaseConnectionMaker`, in its init in lieu of a `database_config_file`.

Given its use as both in CLI and in setup, it should probably be written to be compatible with both a string `database_config_file` and a `DatabaseConnectionMaker` if that's the direction we want to go.

jimmymathews:
I think it is not quite as simple as that. `DatabaseConnectionMaker` and `database_config_file` are not alternative orthogonal ways to do db access, one contains the other. The things that use `database_config_file` but not `DatabaseConnectionMaker` are generally doing so because they exist at a higher abstraction level (CLI scripts that are intermediaries between manual user commands and internal library functions). These will often use `DatabaseConnectionMaker` somewhere down the chain. This is the case in nearly all the source files containing the symbol `database_config_file`:

```txt
spatialprofilingtoolbox/db/credentials.py
spatialprofilingtoolbox/db/schema_infuser.py
spatialprofilingtoolbox/db/stratification_puller.py
spatialprofilingtoolbox/db/scripts/index_expressions_table.py
spatialprofilingtoolbox/db/scripts/create_schema.py
[...]

Issue #120, opened 2023-01-19
New workflow, spatial statistics


jimmymathews:
For this issue, create a new workflow that loads up a given dataset, probably by using the functionality of #76 once completed, then computes spatial statistics functions as provided by the [`squidpy`](https://squidpy.readthedocs.io/en/stable/api.html) package.

This is an exercise in learning the format required by Squidpy functions and transforming what we have into this format. Most likely this format needs to be guessed from the example datasets provided as part of the `squidpy` package, because it is not really documented (beyond the indication that the input format should be in an "AnnData" container).

- The resulting features should get loaded into the `feature_specifier/feature_specification/quantitative_feature_value` system.
- The features should be exposed via a new API endpoint (hopefully not much different from the proximity metric endpoint). This may require a new materialized view, mimicing the way the proximity metrics are made available to the relevant API handler.

(This work is arguably blocked by #83, since #83 should provide a clean template for new workflows.)

(This issue subsumes #41 and #42.)

jimmymathews:
This is being worked on in branch `issue120`. Currently the AnnData object is roughly created in a shell of a new workflow.

jimmymathews:
The "co-occurrence probability" values are basically calculating (in controlled test setup). The return value from squidpy needs interpretation due to lack of doc.

PR #183, opened 2023-07-27
Add squidpy metrics

CarlinLiao:
Per #176, this adds four squidpy metrics, three `ondemand` and one precalculated as part of the `db`, to the suite of features provided by SPT. In doing so, it also does a little refactoring of the Providers classes in `ondemand` to share the same base class `Provider` to reduce redundancy in the way JSON artifacts are loaded (in development code if not in memory).


CarlinLiao:
Outstanding items before merge
* Old tests pass for refactored functions (specifically for `ProximityProvider` and `CountsProvider`)
* The `db` command won't work until some resolution to the problem in #182 is found for the `DatabaseConnectionMaker` to `database_config_file` problem is resolved.)
* Should probably write some new tests for the squidpy functions too

CarlinLiao:
A big problem has emerged while trying to follow the `ProximityProvider` pattern for `SquidpyProvider`: 
* given a tissue sample from a study, a radius length, and pair of phenotype signatures, `ProximityProvider` calculates a single `float`.
* given a tissue sample from a study and one or more phenotype signatures to cluster on, `SquidpyProvider` calculates... a lot of things. The type signature of the return value is `dict[str, dict[str, list[float] | list[int]] | dict[str, list[float]] | dict[str, list[list[float]] | list[float] | list[int]]]`.

I suppose going by the current model each value in that return value should have its own `feature_specification`, but I don't think we should do that. Commit with all this incoming. How to resolve?

CarlinLiao:
I've noticed that in `counts_request_handler`, the proximity calculation's `radius` is an `int`, whereas in other places like `ProximityProvider.retrieve_specifiers` it's a `float`. Is this intentional?

CarlinLiao:
Ready for review @jimmymathews (but not for merge per above outstanding items).

All this refactoring has reinforced my opinion that type hints are Very Good and Very Important.

jimmymathews:
I am still not 100% convinced that it will be possible to refactor the ondemand functions to retrieve from the same in-memory custom data store. Each of these functions works because of special features of their respective in-memory formats. At best it seems we will be able to encapsulate these formats/stores in a single abstraction, but internally they are still separate data structures. I will review your recent commits before commenting further, maybe you have addressed these issues!

jimmymathews:
Yes, after we introduced a bunch of data types a week or a few weeks ago, it is now prudent to start using them. Like `PhenotypeCriteria`, exactly as you say.

jimmymathews:
I don't see a reason not to use `feature_specification` as it was designed, for each feature. If we compute 5 features using the Squidpy library, they can just be documented in that system. The apparently complicated type expression is of course not that complicated when understood as the dysjunction of 5 relatively unrelated types.
Why do you think we should not do this?

spatialprofilingtoolbox/ondemand/providers/squidpy_provider.py
```diff
@@ -0,0 +1,107 @@
...
+        for sample_identifier in sample_identifiers:
+            values = compute_squidpy_metrics_for_one_sample(
+                self.get_cells(sample_identifier, study_name), phenotypes)
+            # TODO: Figure out how to upload vector and matrix squidpy metrics.
+            message = 'Computed feature values of %s: %s, %s'
+            logger.debug(message, feature_specification, sample_identifier, values)
+            with DBCursor() as cursor:
+                for value in values:
+                    add_feature_value(feature_specification,
+                                      sample_identifier, value, cursor)
```
CarlinLiao, line 104: > I don't see a reason not to use `feature_specification` as it was designed, for each feature. If we compute 5 features using the Squidpy library, they can just be documented in that system (feature specification with specifiers etc.). The apparently complicated type expression is of course not that complicated when understood as the disjunction of 5 relatively unrelated types. Why do you think we should not do this?

It's not 5 features, it's more on like 5,000 because it's a matrix. For each sample. And it's not consistent either, it's dependent on how many clusters you choose to have when you input phenotypes. 2 for 1 phenotype, 3 for 2 phenotypes, and so on.

If my understanding is correct, that's thousands of features per sample, each with their own feature specification, from a single study-clustering setup. That seems unwieldy to me with the current setup.

We should either find a new way of storing features, change the current way to accommodate vectors and matrices, or figure out how to boil down the large mass of squidpy return values into one or a few floats.

jimmymathews: I don't think this counts as a feature at all. I would call a 5000-entry matrix computed from source data the result of a data transformation. "Feature" connotes quantitative summary which has a human-understandable interpretation. From the `scstudies` schema:

**[Feature derivation method](https://adiframework.com/docs_site/scstudies_entities_alphabetical.html#feature_derivation_method)**. *The plan for a computational or experimental process that results in associations of quantitative (or, sometimes, qualitative) values to elements of a sample set. Often the associations involve an aggregation over finer-grained samples to a coarsest level of granularity, and are dependent on enumerated specifiers, like an ordered tuple of genes or image channels.*

jimmymathews: I will take a stab at writing the descriptions of the feature derivation method for the discernible outputs of squidpy.

spatialprofilingtoolbox/apiserver/app/main.py
```diff
@@ -9,24 +9,19 @@
...
-from spatialprofilingtoolbox.db.exchange_data_formats.metrics import CellFractionsSummary
-from spatialprofilingtoolbox.db.exchange_data_formats.metrics import PhenotypeSymbol
-from spatialprofilingtoolbox.db.exchange_data_formats.metrics import Channel
-from spatialprofilingtoolbox.db.exchange_data_formats.metrics import PhenotypeCriteria
-from spatialprofilingtoolbox.db.exchange_data_formats.metrics import PhenotypeCounts
-from spatialprofilingtoolbox.db.exchange_data_formats.metrics import \
-    ProximityMetricsComputationResult
+from spatialprofilingtoolbox.db.exchange_data_formats.metrics import CellFractionsSummary, \
+    PhenotypeSymbol, Channel, PhenotypeCriteria, PhenotypeCounts, \
+    ProximityMetricsComputationResult, SquidpyMetricsComputationResult
```
jimmymathews, line 17: Fair warning, I think I'm going to change this back. I wrote up my case for doing multiple imports on their own line (visual scanning stress-reduction). I am open to being convinced to use the mashed version, if you want to make the case.

spatialprofilingtoolbox/apiserver/app/validation.py
```diff
@@ -13,24 +13,29 @@ def abbreviate_string(string: str) -> str:
         abbreviation = abbreviation + '...'
     return abbreviation
 
-async def valid_study_name(study: str=Query(min_length=3)) -> str:
+
+async def valid_study_name(study: str = Query(min_length=3)) -> str:
```
jimmymathews, line 17: Ah, I missed the memo about this case. I thought that the default expressions for keyword arguments expressions were supposed to NOT pad the `=` with spaces according to some style mandate. It turns out that this is true normally but not when the argument is also type hinted (PEP8).
There are many places that I have been using the non-recommended style here.

spatialprofilingtoolbox/apiserver/app/main.py
```diff
@@ -176,20 +174,38 @@ async def request_phenotype_proximity_computation(
...
 
 
+@app.get("/request-squidpy-computation/")
+async def request_squidpy_computation(
+    study: ValidStudy,
+    phenotypes: list[ValidPhenotype],
+) -> SquidpyMetricsComputationResult:
+    """Spatial proximity statistics between phenotype clusters as calculated by Squidpy."""
+    criteria: list[PhenotypeCriteria] = [
+        query().retrieve_signature_of_phenotype(p, study) for p in phenotypes]
```
jimmymathews, line 198: Again, I think I will systematically change the open-closed parens or braces pattern to the one that creates scope blocks that are easily visible without reading the block contents. In this case:
```py
    criteria: list[PhenotypeCriteria] = [
        query().retrieve_signature_of_phenotype(p, study) for p in phenotypes
    ]
```
The point is that, upon seeing/encountering the `[` character, there is a non-recursive algorithm for finding its matching brace `]` that does not require pushing this thought onto a stack and reading the entirety of the enclosed block. Namely, go to the first character of the line (in this case the `c` in criteria), then go down until `]` is encountered.
There is no such algorithm in the alternative form, besides "scan every character until `]` is encountered, unless there are other `[`, in which case increase a mental counter, then scan until a `]` is encountered, unless there are other `[`or `]`, in which case increment or decrement the mental counter accordingly, then scan until ...". This is unnecessary stress in our lives.

jimmymathews: (Obviously, I am disregarding likely conflicts with style mandates. If they do conflict, they probably do not take into account the consideration above, which is more important to me than global style compliance.)

jimmymathews: (This gives me the idea that source code is actually not usually stored in quite the right data structure; source code is really a matrix of characters, not a sequence.)

spatialprofilingtoolbox/apiserver/app/main.py
```diff
@@ -176,20 +174,38 @@ async def request_phenotype_proximity_computation(
...
-            ],
+            ),
         )
     return metrics
 
 
+@app.get("/request-squidpy-computation/")
+async def request_squidpy_computation(
+    study: ValidStudy,
+    phenotypes: list[ValidPhenotype],
```
jimmymathews, line 194: This does indeed generalize the Proximity calculation call signature, but I don't think this is the model we want to emulate in the case of the phenotype arguments. Rather, the model is the Counts calculation, since that supports user-defined (arbitrary) composite phenotype definitions.
If we were limiting ourselves only to the pre-existing phenotype definitions plus single channels, then we really could pre-calculate everything with no need for the ondemand functionality.

jimmymathews: (We would like this for proximity metrics as well, but we just haven't developed the UI.)

spatialprofilingtoolbox/ondemand/providers/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""Classes that provide metric calculation on demand."""
+
+from spatialprofilingtoolbox.ondemand.providers.provider import Provider
+from spatialprofilingtoolbox.ondemand.providers.pending_provider import PendingProvider
+from spatialprofilingtoolbox.ondemand.providers.counts_provider import CountsProvider
+from spatialprofilingtoolbox.ondemand.providers.proximity_provider import ProximityProvider
+from spatialprofilingtoolbox.ondemand.providers.squidpy_provider import SquidpyProvider
```
jimmymathews, line 7: I systematically failed to consider using the `__init__` files to export symbols up the chain of nested modules.
This could be a way to reduce the verbosity and repetition of the single-symbol-per-import-line style that I have been recommending.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
```diff
@@ -0,0 +1,103 @@
+"""Count cells for a specific signature, over the specially-created binary-format index."""
```
jimmymathews, line 1: Hm too bad this was not `git mv`-ed? There is no git diff, it appears as new code.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
```diff
@@ -0,0 +1,103 @@
...
+from spatialprofilingtoolbox.standalone_utilities.log_formats import colorized_logger
+
+logger = colorized_logger(__name__)
+
+
+class CountsProvider(Provider):
+    """Scan binary-format expression matrices for specific signatures."""
+
+    def __init__(self, data_directory: str, load_centroids: bool = False) -> None:
+        """Load from a precomputed JSON artifact in the data directory.
```
jimmymathews, line 15: Is leading triple quotes like this the recommended style?
I thought in blocks you are supposed to bound the docstring above and below the same way, unless it is a single line.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
CarlinLiao: I moved it into the provider subdirectory so it shows up as new. I also excised several functions that overlapped with `ProximityProvider`.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
CarlinLiao: I believe putting the one liner as close to the definition as possible is the recommended style.

jimmymathews: Hm most of the examples in PEP8 and PEP257 have the leading triple quotes. Except that there is maybe an insinuation that the leading triple quotes is more like the "normal form" that docstring processing tools should create, not the way source code should be written. The normal-form-producing code is even provided verbatim in the text of PEP257.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
jimmymathews: I think if you use `git mv` instead of `mv` then git diff will work correctly, like:
```txt
diff --git a/example.md b/example2.md
similarity index 99%
rename from example.md
rename to example2.md
index cfb57be..bac8b60 100644
--- a/example.md
+++ b/example2.md
@@ -1,6 +1,8 @@
 
+New text....
+

```

jimmymathews: But perhaps this only works if you move the file before editing.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
jimmymathews: (I'm talking about the lack of newline after the first `"""`, not distance between a "one liner" and "definition".)

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
```diff
@@ -0,0 +1,194 @@
...
+logger = colorized_logger(__name__)
+
+
+class PendingProvider(Provider, ABC):
+    """Provide precalculated metrics or start, wait, receive, and then provide them."""
+
+    def get_metrics(
+        self,
+        study_name: str,
+        **kwargs: int | PhenotypeCriteria | list[PhenotypeCriteria]
```
jimmymathews, line 23: Amazing, you can type hint arbitrary anonymous keyword arguments? I did not know.

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
```diff
@@ -0,0 +1,194 @@
+"""Abstract class for providers that need to wait for on demand calculations to complete."""
```
jimmymathews, line 1: I am really glad you created this abstract class. It is very simplifying, and I think it would be even if we only had 1 concrete implementation.

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
```diff
@@ -0,0 +1,194 @@
...
+                is_pending = True
+        return self._query_for_computed_feature_values(
+            feature_specification, still_pending=is_pending)
+
+    @abstractmethod
+    @classmethod
+    def get_or_create_feature_specification(
+        cls,
+        study_name: str,
+        **kwargs: int | PhenotypeCriteria | list[PhenotypeCriteria]
```
jimmymathews, line 52: The specifiers for a feature specification can theoretically be quite arbitrary. The explicit list of types here are just parroting back the 3 cases we currently have and are generalizing: radii, a phenotype, or multiple phenotypes. But the `scstudies` schema allows that the specifiers could be arbitrary strings, whose meaning is supposed to be determinable by reading the sentential "feature derivation method" text. I'm thinking about possible specifiers like "**mean** or** median**" in case the feature derivation method involves a statistical center value. I think it is misleading to indicate that these specifiers will always be cell phenotypes or a single numerical quantity.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
CarlinLiao: If I recall correctly, I created a new copy of the file (edited), committed it, and then deleted the old files in a separate commit? Unsure. I think this usually gets picked up correctly as I've seen it before and I've never used `git mv` in my life. It could also be that the changes were so different between the copied, edited version and the original that it didn't get picked up.

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
jimmymathews: So I guess I am advocating `specifiers: tuple[str]`.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
CarlinLiao: I pattern it off the examples and my personal preference to maintain a close physical continuity between the brief description of the method with the method definition and its parameters/typestrings.

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
jimmymathews: I seem to be leading us down a road toward a certain conflict, between:

1. The notion that features that are cell phenotypes should be specifiable on-demand (the so-called anonymous phenotypes, passed by value, as a list of positives and and list of negatives).
2. The notion that feature specifications (in order for the values to be stored correctly in the DB) should have a definite fixed-length list of specifier strings.

The conflict is resolveable easily if part of the on demand job is to insert new phenotypes sometimes, with internal-use-only names that are just integers or something similar.

That proposal raises some questions for me though, about possibly giving too much record-insertion power to arbitrary API users. Maybe it is not such a big deal, after all we are inserting computed values on request.

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
jimmymathews: I'm considering converting all those static methods to object methods and moving all the `feature_specification` arguments into object state as a member variable...

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
jimmymathews: Oh, I think I understand now. I thought by "close" you were referring to proximity between 
```
"""Load from a precomputed JSON artifact in the data directory.
```
and
```
Note: CountsProvider never loads centroids because it does not need them
"""
```
In both forms this proximity is 2 newlines:

```
"""Load from a precomputed JSON artifact in the data directory.

Note: CountsProvider never loads centroids because it does not need them
"""
```

```
"""
Load from a precomputed JSON artifact in the data directory.

Note: CountsProvider never loads centroids because it does not need them
"""
```
So by "one liner" you meant "the (first line of the) docstring" and by "definition" you meant "the call signature expression" (not the method definition).

This reasoning makes sense. I will use this style from now on.

spatialprofilingtoolbox/apiserver/app/main.py
CarlinLiao: Maybe we can compromise on the parenthetical version
```
from spatialprofilingtoolbox.db.exchange_data_formats.metrics import (
    CellFractionsSummary,
    PhenotypeSymbol,
    ...
)
```

jimmymathews: Oh yeah, I was OK with this because it equally well solves the visual scanning problem.

jimmymathews: And supports the easy-line-permuting functionality.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
jimmymathews: Gotcha, that makes sense. I believe I have also experienced git intelligently figuring out that a file has been renamed without explicitly being told about it. After I noticed it fail to do so sometimes, I changed to always using `git mv` for files under source control.

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
CarlinLiao: The important things that we need to keep track of are
* what source data columns are we using (radii and phenotypes, in this case)
* how we transformed them (this is what I think the feature specification is/should be)
* what that transformed data is
for the sake of replicability. (To that end we may also want to track the random seed as well lol.)

`tuple[str]` is fine but if we're going to do that, at least for `SquidpyProvider` and `PhenotypeProvider`, we should extend the function such that `PhenotypeCriteria` can get involved in the typing and use the controlled criteria_to_string function to make it maximally clear what should be done for future Providers. The feature specification should be as formulaic as possible if we're going to let it be a bunch of strings.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
jimmymathews: I pushed a commit implementing this style. I used a short python script to regex replace, it looked right in a sample of examples but I'm not 100% sure every docstring was treated corrrectly by my script.

spatialprofilingtoolbox/apiserver/app/main.py
```diff
@@ -9,24 +9,19 @@
 from fastapi import Response
 from fastapi.responses import StreamingResponse
 
-from spatialprofilingtoolbox.ondemand.counts_service_client import CountRequester
+from spatialprofilingtoolbox.ondemand.service_client import OnDemandRequester
```
jimmymathews, line 12: Really glad we finally have an accurate name for this class.

spatialprofilingtoolbox/db/exchange_data_formats/metrics.py
```diff
@@ -76,17 +71,27 @@ class PhenotypeCounts(BaseModel):
...
+    """Response to an on demand request for computation of Squidpy metrics."""
+
+    values: dict[
+        str,
+        dict[str, list[float] | list[int]] |
+        dict[str, list[float]] |
+        dict[str, list[list[float]] | list[float] | list[int]]
+    ]
+    is_pending: bool
+
```
jimmymathews, line 91: Just a comment: These type definitions in `exchanged_data_formats` are designed specifically for the final API. The priority here is making the API docs readable/understandable by casual browsers, and to make it easier to read/review the API handlers in `app/main.py`.
The defs are in many cases extensively decoupled from the `scstudies` schema, where the priorities are different.
The point of my comment is that more-internal functions (like the providers' logic) may be better suited to types that are not the `exchanged_data_formats` types. This is a case-by-case basis thing.

spatialprofilingtoolbox/ondemand/providers/pending_provider.py
jimmymathews: Note: I wouldn't call radii/phenotypes in this example "source data columns". They are "inputs" but more in the sense of being parameters for a function (namely the feature derivation function), not first-class source data values. (The real source data is anything from the study's original DB instance; cell locations, channel values, channel metadata for the antibodies, diagnosis, intervention, specimen types... .)

Again, I also wouldn't use the term "transformed" here because "transformed" connotes a more general concept than we want. E.g. it allows 5000-entry matrices as the transformed values. We don't want just any type of transformed data to get recorded here, only "features". The verb I've been using is "derive" as in "feature derivation method".

Yes I agree that `PhenotypeCriteria` or similar should be involved somehow in the typing in the implementation of the providers. I think I'm arguing that it should not be involved in the provider interface however, since the exact types are completely feature-type-dependent.
On the other hand the provider interface *should* indicate how to do the object-to-string, as you say.
Perhaps we need `specifiers: tuple[FeatureSpecifier]` and for `PhenotypeCriteria` to implement a `FeatureSpecifier` interface with `to_string`.

spatialprofilingtoolbox/apiserver/app/validation.py
```diff
@@ -39,11 +44,10 @@ def valid_single_or_composite_identifier(identifier) -> str:
...
     raise ValueError(f'Channel name or phenotype identifier invalid: {abbreviation}')
 
-async def valid_phenotype1(phenotype1: str=Query(min_length=1)) -> str:
-    return valid_single_or_composite_identifier(phenotype1)
 
-async def valid_phenotype2(phenotype2: str=Query(min_length=1)) -> str:
-    return valid_single_or_composite_identifier(phenotype2)
+async def valid_phenotype(phenotype: str = Query(min_length=1)) -> str:
+    return valid_single_or_composite_identifier(phenotype)
+
```
jimmymathews, line 50: Unfortunately, FastAPI seems to pick up the attribute names verbatim and use them as the query parameter names, so I don't think that replacing `phenotype1` and `phenotype2` with `phenotype` in the `ValidPhenotype` works. I am getting errors about "phenotype" field missing, when "phenotype1" and "phenotype2" are supplied, even though they are called `phenotype1` and `phenotype2` in the call signature for `request_phenotype_proximity_computation`.
The workaround was to created separate types with the respective names (`ValidPhenotype1` and `ValidPhenotype2`).

spatialprofilingtoolbox/ondemand/request_handler.py
```diff
@@ -0,0 +1,186 @@
...
+            raise exception
+
+        message = dumps(metrics) + self._get_end_of_transmission()
+        self.request.sendall(message.encode('utf-8'))
+        return True
+
+    def _get_proximity_metrics(self, study, radius, signature):
+        positives1, negatives1, positives2, negatives2 = signature
+        phenotype1 = PhenotypeCriteria(positives1, negatives1)
+        phenotype2 = PhenotypeCriteria(positives2, negatives2)
```
jimmymathews, line 54: One of the downsides of Pydantic as compared with `attrs` is that this constructor format is not supported. In Pydantic you have to provide the controlled attributes by keyword.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
```diff
@@ -0,0 +1,103 @@
...
+    ) -> dict[str, list[int]]:
+        """Count the number of structures that part of this signature, in signs."""
+        counts = {}
+        for specimen, data_array in self.data_arrays[study_name].items():
+            count = 0
+            for entry in data_array['entry']:
+                if (entry | positives_signature == entry) and \
+                        (~entry | negatives_signature == ~entry):
+                    count = count + 1
+            counts[specimen] = [count, len(data_array)]
```
jimmymathews, line 80: This is always [0, 0] in tests.

spatialprofilingtoolbox/apiserver/app/validation.py
CarlinLiao: FastAPI more like SlowAndOvercomplicatedAPI

spatialprofilingtoolbox/ondemand/request_handler.py
CarlinLiao: We know how to keyword type now so that's less of an issue.

spatialprofilingtoolbox/ondemand/providers/proximity_provider.py
```diff
@@ -5,13 +5,19 @@
 
 from spatialprofilingtoolbox.db.database_connection import DBCursor
 from spatialprofilingtoolbox.db.exchange_data_formats.metrics import PhenotypeCriteria
-from spatialprofilingtoolbox.ondemand.phenotype_str import \
-    phenotype_str_to_phenotype, phenotype_to_phenotype_str
+from spatialprofilingtoolbox.ondemand.phenotype_str import (\
```
CarlinLiao, line 8: Slash endlines aren't necessary with the parenthetical wrap.

spatialprofilingtoolbox/ondemand/providers/counts_provider.py
jimmymathews: I think this line is the culprit:

https://github.com/nadeemlab/SPT/blob/d38166e5a1a5e11d8ffdbc486e74a8db87bb6d53/spatialprofilingtoolbox/ondemand/providers/provider.py#L105

It doubly reads into the buffer, in one of the two cases only.

jimmymathews: I'm workin on it.

spatialprofilingtoolbox/workflow/common/proximity.py
```diff
@@ -1,41 +1,25 @@
...
-        range1 = [False,]*(loc.start - 0)
-        range2 = [True,]*(loc.stop - loc.start)
-        range3 = [False,]*(cells.shape[0] - loc.stop)
-        return np.asarray(range1 + range2 + range3)
-    if isinstance(loc, int):
-        return np.asarray([i == loc for i in range(cells.shape[0])])
-    raise ValueError(f'Could not select by index: {multiindex}. Got: {loc}')
-
-def compute_proximity_metric_for_signature_pair(signature1, signature2, radius, cells, tree):
+def compute_proximity_metric_for_signature_pair(
```
jimmymathews, line 16: This change broke the function's signature, there is another usage in proximity/core.py. I'm updating.

jimmymathews: Note that it is a bit against the design intention (as remarked elsewhere) to start using the `exchange_data_formats` pydantic types for other internal uses, i.e. for uses other than exchange via the API. Not sure what to recommend as alternative right now, so this is OK.

spatialprofilingtoolbox/workflow/phenotype_proximity/core.py
```diff
@@ -3,19 +3,23 @@
 """
 import warnings
 import pickle
+from typing import cast
 
 import pandas as pd
-from sklearn.neighbors import BallTree
+from pandas import DataFrame
+from sklearn.neighbors import BallTree # type: ignore
```
CarlinLiao, line 10: Linters prefer two spaces before inline comments

jimmymathews: Ok, fixed. I wonder why it wasn't caught (using MyPy, Pylint, Pylance). Maybe my pylint isn't active somehow.

CarlinLiao: autopep8 usually does this for me

spatialprofilingtoolbox/workflow/common/proximity.py
CarlinLiao: Why wouldn't we want to use the exchange data formats wherever possible? The controlled types make it easier for us to understand and pass complete information as well.

jimmymathews: Clearly controlled types make it easier for us to understand and pass complete information.
I did not recommend uncontrolled types.
I am recommending against using the types, which were designed/selected/documented specifically for the web-application-supporting API, in the internal operation of the library. In my opinion, doing so would be an example of the anti-pattern of needlessly increasing coupling between unrelated modules.

jimmymathews: The kind of things I'm worried about is attributes of the exchange data types module having surprise effects on internal functions, like the bug at the top of this thread that spawned this conversation.

jimmymathews:
All tests now pass! Now moving on to some minor linting tasks and other minor changes. Then will add tests for the squidpy provider.

jimmymathews:
I have completed my pass of style conformation and addressing linter complaints, everywhere except in the new, squidpy-related modules. There are several things I want to do in these modules, which I'll note line by line.

spatialprofilingtoolbox/workflow/common/squidpy.py
```diff
@@ -0,0 +1,97 @@
...
+
+from spatialprofilingtoolbox.db.exchange_data_formats.metrics import PhenotypeCriteria
+from spatialprofilingtoolbox.workflow.common.cell_df_indexer import get_mask
+
+
+def describe_squidpy_feature_derivation_method():
+    return '''
+    Calculates nhood_enrichment, co_occurrence, and ripley from squidpy.gr using clusters derived
+    from the phenotypes provided. Reference db.squidpy_metrics.convert_df_to_anndata for the
+    clustering method.
```
jimmymathews, line 18: This is not meant to be an uncontrolled description string, it is supposed to exactly tell what feature values were intended to be computed, given the specific definite list of specifiers for the feature. E.g.:

```txt
For a given cell phenotype (first specifier), the average number of cells of a second phenotype (second specifier) within a specified radius (third specifier).
```
In the example, there are exactly 3 specifiers, and the description tells what was going on in real life if you get a given number as the feature value for a given sample.
This description becomes a data value in the database, not a docstring somewhere to be read by a developer. So references to implementation `squidpy.gr` and `db.squidpy_metrics...` are also against what was intended.
It seems there are 3 or 4 different things to write here, depending on which specific feature class is computed (Ripley, co-occurence, etc.).
We should write these after we are sure about which thing is the final feature value actually recorded. As discussed in another thread, a 5000-entry matrix is not an acceptable "value" for a summary feature, so there is some more work to do here.

spatialprofilingtoolbox/apiserver/app/validation.py
```diff
@@ -75,6 +74,16 @@ async def valid_channel_list_positives(positive_marker: ChannelList) -> list[str
 async def valid_channel_list_negatives(negative_marker: ChannelList) -> list[str]:
     return valid_channel_list(negative_marker)
 
+
+squidpy_feature_classnames = ('neighorhood enrichment', 'co-occurrence', 'ripley')
```
CarlinLiao: Are you only allowing the user to retrieve one of the three Squidpy features at a time? The three of them share the initialized neighborhood calculation that would have to be repeated if it isn't saved, and I don't see a reason why a user would prefer just one instead of all three of them, since they share the same input parameters. (Although it could be beneficial to return them sequentially as they're finished calculating.)

jimmymathews: Hm OK, I didn't think about how that overhead might be a dominant expenditure of computation time.
This will make the squidpy metrics break the mold a little, providing multiple features and computing multiple features on every query.

jimmymathews: There are numerous places where one occurrence of query handling is associated with one feature specification, so these will have to be rewritten. For exapmle `PendingProvider.get_metrics`, which was supposed to be the generalized implementation.

jimmymathews: I discussed this a little in the other threads:
https://github.com/nadeemlab/SPT/pull/183#discussion_r1287726451 (*... supposed to exactly tell what feature values were intended to be computed ...*)
https://github.com/nadeemlab/SPT/pull/183#discussion_r1279753401 (*...I don't think this counts as a feature at all.*)

I have been sticking to the original planned goals of #120, including:

*The resulting features should get loaded into the* `feature_specifier/feature_specification/quantitative_feature_value` *system.*

The non-scalar return values you are referring to are not yet summarized, so they are not first-class features according to my "rigid" definition, which requires the simple/univariate aspect.

But that is not all. My definition also stipulates that the derivation method should be carefully described, and we still lack this as well.

My strategy for achieving explainable outputs is (1) identify representative summary features for each unstructured bundle, (2) research the implementation to understand what the summary feature of the unstructured bundle means.

This is what I plan to do after refactoring the squidpy-related modules in order to be feature-oriented.

jimmymathews: This turned out to be pretty easy for the neighborhood enrichment test.
Here is an example of the return value:
```json
 {
    "zscore": [
        [0.5406386960248688, -0.19020676295202088, 0.4386328402677137],
        [-0.3693187058572753, 0.3503620265496884, -0.4386328402677137],
        [-1.0974813402382708, -0.992579671981139, null]
    ],
    "count": [
        [104, 142, 3],
        [141, 204, 3],
        [1, 2, 0]
    ]
}
```
I checked the source code, and here is my draft description:

```txt
For two given cell phenotypes (first and second specifier), the
z-score of the number of occurrences of neighbor relations
between two cells of the given phenotypes, the z-score computed
in reference to a bootstrapped empirical distribution constructed
using 1000 random permutations of the cell set with labels fixed.
```

The "non-scalar" return value in this case is mainly non-scalar because we have not restricted our attention to the 2 phenotypes which are the main arguments. (The matrix of z-scores is a z-score for each pair of phenotypes). For our purposes I think we should convert this z-score back to a probability value (and change "z score" => "p value" in the description), for a more direct interpretation of the result: (an esimate of) the fraction of the number of times that the random bootstrapping value was higher than the observed count.

In addition to z-scores, "counts" are returned. This is essentially the "test statistic" value corresponding to the p-value. I don't think it is essential to record this and report it in the response. If it turns out to be more important than I thought, we can always make a new feature for it.

jimmymathews: Note that to get the actual fraction (as opposed to "an estimate of"), we would need to slightly change the squidpy implementation. Specifically this line:

https://github.com/scverse/squidpy/blob/8cd23551e5e14a95ae769d426c33813e5d889179/src/squidpy/gr/_nhood.py#L188C5-L188C62

```py
zscore = (count - perms.mean(axis=0)) / perms.std(axis=0)
```

It would need to be something like
```py
fraction = sum([1 for _count in perms if _count > count]) / len(perms)
```

CarlinLiao: Are you suggesting we fork and maintain a separate branch of squidpy for SPT?

jimmymathews: The estimate is good enough for our purposes, but when we are ready we could make a PR and request a feature that makes this information available optionally.

jimmymathews: I made a first pass at the same task for the co-occurrence scores. Even after studying the squidpy implementation, I don't have perfect confidence that I have understood the outputs correctly. I tested it with varying parameters (mainly `interval`), and I have a reasonable theory about the output format. I think it is a 3-axis array, indexable respectively by: cluster, cluster, and distance interval. If they are ordered as expected, I think I am extracting the correct value.
As implemented now, this metric depends on 3 specifiers: a phenotype, a second phenotype, and a radius. I had to add a radius parameter throughout, since previously we assumed that only a phenotype list would be required to specify one of these features.
At the UI level we will likely use a reasonable default for this radius value, but I think it should still be configurable by the (power) user.

spatialprofilingtoolbox/ondemand/request_handler.py
```diff
@@ -147,15 +147,27 @@ def _get_phenotype_pair_specification(self, groups):
...
+        match feature_class:
+            case 'neighborhood enrichment':
+                if len(groups) != 4 + 1:
+                    return False
+            case 'co-occurrence':
+                if len(groups) != 4 + 1 + 1:
+                    return False
+            case 'ripley':
+                if len(groups) != 2 + 1:
+                    return False
```
CarlinLiao, line 159: Why do neighborhood and ripley have different group lengths?

spatialprofilingtoolbox/workflow/common/squidpy.py
```diff
@@ -39,22 +44,33 @@ def compute_squidpy_metric_for_one_sample(
...
+    return float(norm.cdf(zscore))
 
 
 def summarize_co_occurrence(unstructured_metrics) -> float | None:
-    return 2
+    occurrence_ratios = unstructured_metrics['occ']
+    return float(occurrence_ratios[0][1][0])
 
 
 def summarize_ripley(unstructured_metrics) -> float | None:
```
CarlinLiao, line 65: These should be private methods.

spatialprofilingtoolbox/ondemand/request_handler.py
jimmymathews: The neighborhood enrichment metric depends on a focal phenotype, and another phenotype that is the one considered for being the "riches" that the former phenotype's cells are either "enriched in" or not. The Ripley statistic depends on a single point set, not a pair.

spatialprofilingtoolbox/workflow/tabular_import/parsing/sample_stratification.py
```diff
@@ -1,5 +1,7 @@
 """Source file parsing regarding sample-level cohort identification."""
 import re
+from typing import Callable
+from typing import Any
```
CarlinLiao, line 4: ,

jimmymathews: If there are only two symbols imported, I'm not worried about creating an enumeration of them.

spatialprofilingtoolbox/db/scripts/list_studies.py
```diff
@@ -5,9 +5,10 @@
 from spatialprofilingtoolbox.workflow.common.cli_arguments import add_argument
 from spatialprofilingtoolbox.standalone_utilities.module_load_error import SuggestExtrasException
 try:
-    from spatialprofilingtoolbox import DatabaseConnectionMaker  # pylint: disable=ungrouped-imports
+    from spatialprofilingtoolbox import DatabaseConnectionMaker
 except ModuleNotFoundError as e:
     SuggestExtrasException(e, 'db')
+from spatialprofilingtoolbox import DatabaseConnectionMaker
```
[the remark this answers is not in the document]
jimmymathews: So the way it was supposed to work is that the CLI tools for different major submodules (`db`, `workflow`, etc.) were supposed to tell a helpful suggestion in case they are used without the corresponding `pip install spatialprofilingtoolbox[<module name>]`, i.e. if some dependencies are missing. Rather than just triggering the normal unhelpful "import fails" error.
This functionality was sort of tricky to implement without creating linting errors. If the import is in a try catch block, later blocks would complain that the symbol might not be bound.
Today I discovered that I can just use the try block system, then immediately import like normal (not in the try block). This satisfies the later code blocks that use the imported symbols, but doesn't break the "suggestion" functionality.

spatialprofilingtoolbox/ondemand/fast_cache_assessor.py
```diff
@@ -1,10 +1,12 @@
 """Assesses presence of "fast cache" files, and creates/deletes as necessary."""
+
 from os import system
 from os import environ
 from os import remove
 from os import listdir
 from os.path import isfile
 from os.path import join
+from os.path import isdir
```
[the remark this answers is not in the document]
jimmymathews: I added an extra import for a (security motivated) check on the directory existence. I just didn't do the work to modify the import list to use our new convention.

spatialprofilingtoolbox/ondemand/service_client.py
```diff
@@ -129,7 +129,8 @@ def get_squidpy_metrics(
         radius: float | None = None,
     ) -> SquidpyMetricsComputationResult:
         """Get spatial proximity statistics between phenotype clusters as calculated by Squidpy."""
-        assert len(signature) % 2 == 0
+        if not len(signature) in [1, 2]:
```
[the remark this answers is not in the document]
jimmymathews: Fixed

spatialprofilingtoolbox/standalone_utilities/configuration_settings.py
```diff
@@ -1,6 +1,12 @@
 """Configuration settings."""
 from importlib.metadata import version
-
+from importlib.metadata import PackageNotFoundError
```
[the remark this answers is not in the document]
jimmymathews: I'm not doing this for cases with only 2 lines.

spatialprofilingtoolbox/workflow/common/sparse_matrix_puller.py
```diff
@@ -1,5 +1,4 @@
-"""
-Retrieve the "feature matrix" for a given study from the database, and store it in a special
+"""Retrieve the "feature matrix" for a given study from the database, and store it in a special
 (in-memory) binary compressed format.
```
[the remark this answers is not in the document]
jimmymathews: I use 100 line length, not sure if that is related.

Wait so the initial syntax with triple quotes and no newline is only for one-line docstrings? I thought the conclusion from the previous discussion was that one never uses "triple quotes then newline" as the initial token for a docstring, always "triple quotes then the contents".

spatialprofilingtoolbox/workflow/scripts/configure.py
```diff
@@ -11,9 +11,19 @@
...
 from importlib.resources import files
 
-from spatialprofilingtoolbox.workflow.common.cli_arguments import add_argument
+try:
+    import jinja2
+except ModuleNotFoundError as e:
+    from spatialprofilingtoolbox.standalone_utilities.module_load_error import \
+        SuggestExtrasException
+    SuggestExtrasException(e, 'workflow')
+import jinja2  # pylint: disable=ungrouped-imports
```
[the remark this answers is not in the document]
jimmymathews: See the other comment about this.

spatialprofilingtoolbox/workflow/tabular_import/parsing/sample_stratification.py
jimmymathews: A syntax that is dis-preferred is "from typing import Callable, Any". The only other syntax I would consider here is:
```py
from typing import (
    Callable,
    Any,
)
```
but that seems overkill for 2 items.

jimmymathews:
I did a broad style conformity / linting / type-hinting exercise over the whole codebase, because this PR seems to touch many source files. Some of the things were unrelated to the PR but were long overdue.
Some things were noticed but still unfixed, for the sake of time.

spatialprofilingtoolbox/db/scripts/list_studies.py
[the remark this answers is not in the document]
CarlinLiao: Interesting. Why wouldn't it just throw the uncaught import error in practice?

CarlinLiao:
Who knew adding a couple of squidpy features would require such big rewrites?

jimmymathews:
Well the bulk of the new changes are not squidpy specific, it is instead the generalization of the on-demand services. This is a very welcome update so I'm not complaining!

spatialprofilingtoolbox/db/scripts/list_studies.py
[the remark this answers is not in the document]
jimmymathews: I looked a little closer, and apparently the way I implemented it was for the original exception to be re-raised. So the ugly error still happens, just after a helpful message.
(The ugly error happens inside SuggestExtrasException's init, so the second occurrence of the import line isn't actually running in such cases.)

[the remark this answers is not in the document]
jimmymathews: Here's how it looks (example: `spt db status` with pandas missing):

image

spatialprofilingtoolbox/db/squidpy_metrics.py
```diff
@@ -18,10 +19,10 @@
 logger = colorized_logger(__name__)
 
 
-def _describe_spatial_autocorr_derivation_method() -> str:
-    """Return a description of the spatial autocorrelation derivation method."""
-    return 'Global Autocorrelation Statistic (Moran\'s I). See Squidpy documentation for ' \
-        'squidpy.gr.spatial_autocorr for more information.'.lstrip().rstrip()
+def describe_spatial_autocorr_derivation_method() -> str:
```
CarlinLiao, line 22: This function was changed to a public method but had its docstring removed in the process?

jimmymathews: The function practically is a docstring...
In general I have eschewed function docstrings in this codebase in favor of semantic symbol names. The other thing I'm doing instead of docstrings is a pattern of frequent minting of new small functions in order to provide yet more names.
In spt v0 there were copious function docstrings, with every argument annotated. But it seemed terribly time-consuming to update the docstrings on every code change, and it was hard to read the code itself when so much of the screen was docstring. I scrapped them all and resolved to write docstrings when the library is done.

This function is made public because library users should be able to access it in order to select specifically this feature from a given dataset, as illustrated by the test case.

jimmymathews:
A test of the Moran I statistic computation has been added.

I think it will make sense to put the implementation of the Moran and the other metrics together, not split them like they are now. Then the module that computes things without user selections can do so for all the metrics that require only 1 phenotype (using the list of channels plus the list of phenotypes). So that's Moran autocorrelation but also the Ripley.

jimmymathews:
All tests pass for me now, including the new tests for the squidpy metrics both (1) as provided via the API, and (2) provided in a less-interactive manner meant for immediately after data import.

jimmymathews (approved):

PR #184, opened 2023-07-27
Normalize database config file vs cursor etc

jimmymathews:
Addresses #182 .

### DIFF
--- a/spatialprofilingtoolbox/cggnn/scripts/run.py
+++ b/spatialprofilingtoolbox/cggnn/scripts/run.py
@@ -164,8 +164,8 @@ def save_importances(_args):
 
 if __name__ == "__main__":
     args = parse_arguments()
-    study_data: Dict[str, Dict] = FeatureMatrixExtractor.extract(
-        args.spt_db_config_location)[args.study]
+    extractor = FeatureMatrixExtractor(database_config_file=args.spt_db_config_location)
+    study_data: Dict[str, Dict] = extractor.extract(study=args.study)
 
     df_cell = _create_cell_df(
         {slide: data['dataframe']
--- a/spatialprofilingtoolbox/cggnn/scripts/run.py
+++ b/spatialprofilingtoolbox/cggnn/scripts/run.py
@@ -1,6 +1,5 @@
 "Run through the entire SPT CG-GNN pipeline using a local db config."
 from argparse import ArgumentParser
-from typing import Dict, Tuple
 from os.path import join
 
 from pandas import DataFrame
@@ -121,7 +120,7 @@ def parse_arguments():
     return parser.parse_args()
 
 
-def _create_cell_df(cell_dfs: Dict[str, DataFrame], feature_names: Dict[str, str]) -> DataFrame:
+def _create_cell_df(cell_dfs: dict[str, DataFrame], feature_names: dict[str, str]) -> DataFrame:
     "Find chemical species, phenotypes, and locations and merge into a DataFrame."
 
     for specimen, df_specimen in cell_dfs.items():
@@ -140,7 +139,7 @@ def _create_cell_df(cell_dfs: Dict[str, DataFrame], feature_names: Dict[str, str
 
 
 def _create_label_df(df_assignments: DataFrame,
-                     df_strata: DataFrame) -> Tuple[DataFrame, Dict[int, str]]:
+                     df_strata: DataFrame) -> tuple[DataFrame, dict[int, str]]:
     """Get slide-level results."""
     df = merge(df_assignments, df_strata, on='stratum identifier', how='left')[
         ['specimen', 'subject diagnosed result']].rename(
@@ -165,7 +164,7 @@ def save_importances(_args):
 if __name__ == "__main__":
     args = parse_arguments()
     extractor = FeatureMatrixExtractor(database_config_file=args.spt_db_config_location)
-    study_data: Dict[str, Dict] = extractor.extract(study=args.study)
+    study_data: dict[str, dict] = extractor.extract(study=args.study)
 
     df_cell = _create_cell_df(
         {slide: data['dataframe']
--- a/spatialprofilingtoolbox/db/feature_matrix_extractor.py
+++ b/spatialprofilingtoolbox/db/feature_matrix_extractor.py
@@ -40,19 +40,21 @@ class FeatureMatrixExtractor:
     ):
         self.cursor = cursor
         self.database_config_file = database_config_file
-        if cursor is None and database_config_file is None:
-            logger.error('Must supply either cursor or database_config_file.')
-            return
-        if cursor is not None and database_config_file is not None:
-            message = 'A cursor and database configuration file were both specified. Using the '\
-                'cursor.'
-            logger.warning(message)
         if cursor is not None:
             self.db_source = DBSource.CURSOR
         elif database_config_file is not None:
             self.db_source = DBSource.CONFIG_FILE
         else:
             self.db_source = DBSource.UNKNOWN
+        self._report_on_arguments()
+
+    def _report_on_arguments(self):
+        if self.cursor is None and self.database_config_file is None:
+            logger.error('Must supply either cursor or database_config_file.')
+        if self.cursor is not None and self.database_config_file is not None:
+            message = 'A cursor and database configuration file were both specified. Using the '\
+                'cursor.'
+            logger.warning(message)
 
     def extract(self,
         specimen: str | None=None,
--- a/spatialprofilingtoolbox/db/feature_matrix_extractor.py
+++ b/spatialprofilingtoolbox/db/feature_matrix_extractor.py
@@ -63,23 +63,24 @@ class FeatureMatrixExtractor:
         continuous_also: bool=False,
     ):
         extraction = None
-        if self.db_source == DBSource.CURSOR:
-            extraction = self._extract(
-                specimen=specimen,
-                study=study,
-                continuous_also=continuous_also,
-            )
-        if self.db_source == DBSource.CONFIG_FILE:
-            with DatabaseConnectionMaker(self.database_config_file) as dcm:
-                with dcm.get_connection().cursor() as cursor:
-                    self.cursor = cursor
-                    extraction = self._extract(
-                        specimen=specimen,
-                        study=study,
-                        continuous_also=continuous_also,
-                    )
-        if self.db_source == DBSource.UNKNOWN:
-            logger.error('The database source can not be determined.')
+        match self.db_source:
+            case DBSource.CURSOR:
+                extraction = self._extract(
+                    specimen=specimen,
+                    study=study,
+                    continuous_also=continuous_also,
+                )
+            case DBSource.CONFIG_FILE:
+                with DatabaseConnectionMaker(self.database_config_file) as dcm:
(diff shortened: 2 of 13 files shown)


spatialprofilingtoolbox/cggnn/scripts/run.py
```diff
@@ -164,8 +164,8 @@ def save_importances(_args):
 
 if __name__ == "__main__":
     args = parse_arguments()
-    study_data: Dict[str, Dict] = FeatureMatrixExtractor.extract(
-        args.spt_db_config_location)[args.study]
+    extractor = FeatureMatrixExtractor(database_config_file=args.spt_db_config_location)
+    study_data: Dict[str, Dict] = extractor.extract(study=args.study)
```
CarlinLiao, line 168: No need to use the capital D `Dict` in Python 3.11.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
```diff
@@ -18,24 +18,61 @@
...
+    ):
+        if cursor is None and database_config_file is None:
+            logger.error('Must supply either cursor or database_config_file.')
+            return
+        if cursor is not None and database_config_file is not None:
+            message = 'A cursor and database configuration file were both specified. Using the '\
+                'cursor.'
+            logger.warning(message)
+        self.cursor = cursor
+        self.database_config_file = database_config_file
```
CarlinLiao, line 39: The warning says that the cursor will be used if both cursor and config are provided, but it saves both anyway.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
```diff
@@ -18,24 +18,61 @@
...
+
+    def extract(self,
+        specimen: str | None=None,
+        study: str | None=None,
+        continuous_also: bool=False,
+    ):
```
CarlinLiao, line 45: Missing type hint for output.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
```diff
@@ -18,24 +18,61 @@
...
+        extraction = None
+        if self.cursor is not None:
+            extraction = self._extract(specimen=specimen, study=study, continuous_also=continuous_also)
+        if self.database_config_file is not None:
```
CarlinLiao, line 49: Shouldn't this be an `elif`? (Although it wouldn't be an issue if only one of `cursor` or `database_config_file` were saved in the init.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
```diff
@@ -1,10 +1,10 @@
 """
-Convenience provision of a feature matrix for each study, the data retrieved
-from the SPT database.
+Convenience provision of a feature matrix for each study, the data retrieved from the SPT database.
 """
```
CarlinLiao, line 3: pydocstyle says module docstrings should be one liners if it fits on one line (include the start and end quotes). If not, there should be a short description that fits on one line and elaboration after two line breaks.

spatialprofilingtoolbox/db/stratification_puller.py
```diff
@@ -1,67 +1,79 @@
...
         self.stratification = None
 
-    def pull(self):
-        self.stratification = self.retrieve_stratification()
+    def pull(self) -> None:
+        self.stratification = self._retrieve_stratification()
 
-    def get_stratification(self):
-        return self.stratification
+    def get_stratification(self) -> dict:
```
CarlinLiao, line 25: Can we provide a more detailed type hint for the dict? Looks like it's a `dict[str, dict[str, DataFrame]]`.

spatialprofilingtoolbox/workflow/common/sparse_matrix_puller.py
```diff
@@ -161,61 +170,57 @@ def fill_data_arrays_for_study(
             progress_reporter.increment(iteration_details=_specimen)
         progress_reporter.done()
 
-    def get_pertinent_specimens(self, study_name, specimen: str=None):
+    def _get_pertinent_specimens(self, study_name: str, specimen: str | None=None):
```
CarlinLiao, line 173: This could use a return type hint, ` -> list[str]`.

spatialprofilingtoolbox/workflow/common/sparse_matrix_puller.py
```diff
@@ -161,61 +170,57 @@ def fill_data_arrays_for_study(
...
+        FROM specimen_data_measurement_process sdmp
+        WHERE sdmp.study=%s
+        ORDER BY sdmp.specimen
+        ;
+        ''', (study_name,))
+        rows = self.cursor.fetchall()
         return [row[0] for row in rows]
 
-    def get_study_names(self, connection, study=None):
+    def _get_study_names(self, study: str | None=None) -> list[str]:
```
CarlinLiao, line 186: Why did this internal function get a return type hint when most pubic functions didn't?

CarlinLiao (changes requested):
All tests pass but there are a few clerical notes I think should be addressed before merging.

spatialprofilingtoolbox/cggnn/scripts/run.py
jimmymathews: Yup, fixed.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
jimmymathews: Yes I went back and forth with a few options here.
Originally I wanted the cursor setup to take place immediately, here, with no need for a `database_config_file` attribute. But the context manager business frustrated this, as usual.
The expected behavior is for `cursor` to get used if provided, otherwise use `database_config_file` if available.
It is helpful to store `database_config_file` in `self.database_config_file` even when it is `None`, because it tells that no `database_config_file` was supplied. To make this transparent we could have some additional state variable like `db_source: DBSource`, with enum `DBSource` members `CURSOR` or `CONFIG_FILE`. But that seems like overkill.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
jimmymathews: That's because the output "type" is very complicated. I don't think every final member is a dataframe, as you have supposed.

jimmymathews: This would of course benefit from refactoring to have clearer return types, but that's not really the point of this PR.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
jimmymathews: Yes this is a mistake. The IDE suggested changing `elif` to `if` and I complied, but that was when those `extract` lines were return statements. In this form `elif` is right.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
jimmymathews: Hm yes I've thought about this sometimes. There is an 8 character ambiguity, since making a one-line comment block into one comment line increases the size of the line by the size of `"""  """`. In this case, this goes beyond the 100 character line limit I've been using.
It seems that the pydocstyle prescription asks to alter the content of the docstring in this case. This seems like arbitrary carelessness on the part of the style guide, since it follows that docstrings that are between 93 and 100 characters long are not allowed.

spatialprofilingtoolbox/db/stratification_puller.py
jimmymathews: I think this is a good idea after refactoring so that the return type is actually controlled. But that is not really part of this PR. We could make a new issue if it is important.

spatialprofilingtoolbox/workflow/common/sparse_matrix_puller.py
jimmymathews: Yes, I've also changed this a tuple.

spatialprofilingtoolbox/workflow/common/sparse_matrix_puller.py
jimmymathews: Yes they should have them, I added a few more type hints.
The answer is that this module was written without regard to type hinting. Every type hint was added after the module was written.

spatialprofilingtoolbox/db/stratification_puller.py
jimmymathews: Nevermind, my comment applies to the other pullers not this one. You are right, this one has all final values dataframes.

jimmymathews: (and the nesting levels are all consistent)

spatialprofilingtoolbox/db/feature_matrix_extractor.py
jimmymathews: Well overkill or not I implemented it, and it seems cleaner.

jimmymathews:
I think everything is fixed up.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
```diff
@@ -16,42 +19,67 @@
...
-        if self.cursor is not None:
-            extraction = self._extract(specimen=specimen, study=study, continuous_also=continuous_also)
-        if self.database_config_file is not None:
+        if self.db_source == DBSource.CURSOR:
+            extraction = self._extract(
+                specimen=specimen,
+                study=study,
+                continuous_also=continuous_also,
+            )
+        if self.db_source == DBSource.CONFIG_FILE:
```
CarlinLiao, line 72: Not a blocker due to the new way it's implemented, but this should still be an `elif`, right?

CarlinLiao (approved):

spatialprofilingtoolbox/db/feature_matrix_extractor.py
jimmymathews: I'm not sure, I've seen it recommended to use a sequence of bare `if` statements when exhaustively handling an enum value. I guess for symmetry.
The behavior is theoretically different, since the enum value (`db_source` in this case) could accidentally be altered during one of the handlers...

jimmymathews: (Basically, Python lacks `switch`).

spatialprofilingtoolbox/db/feature_matrix_extractor.py
CarlinLiao: I acquiesce and reword to make it more concise, putting further detail in the body if it calls for it. Your effective short description limit is about 10 characters shorter than your chosen line limit setting.

spatialprofilingtoolbox/db/feature_matrix_extractor.py
jimmymathews: Wait it doesn't. I just saw [this](https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching), apparently availabile since 3.10.

jimmymathews: Updated to use `match case`.

CarlinLiao:
Final change approved.

PR #185, opened 2023-08-08
create cg-gnn extraction script

CarlinLiao:
This creates a `spt cggnn extract` command that pulls information `cg-gnn` needs from a SPT db instance and saves it to two DataFrames and a JSON.

### DIFF
--- a/spatialprofilingtoolbox/cggnn/extract.py
+++ b/spatialprofilingtoolbox/cggnn/extract.py
@@ -1,0 +1,57 @@
+"""Extract information cg-gnn needs from SPT."""
+
+from typing import Any
+
+from pandas import DataFrame, concat, merge  # type: ignore
+from numpy import sort  # type: ignore
+
+from spatialprofilingtoolbox.db.feature_matrix_extractor import FeatureMatrixExtractor
+
+
+def _create_cell_df(
+    dfs_by_specimen: dict[str, DataFrame],
+    feature_names: dict[str, str],
+) -> DataFrame:
+    """Find simple and complex phenotypes, and locations and merge into a DataFrame."""
+    feature_ids_to_names = {ft_id: 'FT_' + ft_name for ft_id, ft_name in feature_names.items()}
+    for specimen, df_specimen in dfs_by_specimen.items():
+        df_specimen.rename(feature_ids_to_names, axis=1, inplace=True)
+        # TODO: Create complex phenotype columns
+        df_specimen.rename({'pixel x': 'center_x', 'pixel y': 'center_y'},
+                           axis=1, inplace=True)
+        df_specimen['specimen'] = specimen
+
+    # TODO: Reorder so that it's simple phenotype, specimen, complex phenotype, x, y
+    df = concat(dfs_by_specimen.values(), axis=0)
+    df.index.name = 'histological_structure'
+    return df
+
+
+def _create_label_df(
+    df_assignments: DataFrame,
+    df_strata: DataFrame,
+) -> tuple[DataFrame, dict[int, str]]:
+    """Get slide-level results."""
+    df = merge(df_assignments, df_strata, on='stratum identifier', how='left')[
+        ['specimen', 'subject diagnosed result']].rename(
+        {'specimen': 'slide', 'subject diagnosed result': 'result'}, axis=1)
+    label_to_result = dict(enumerate(sort(df['result'].unique())))
+    return df.replace({res: i for i, res in label_to_result.items()}), label_to_result
+
+
+def extract_cggnn_data(
+    spt_db_config_location: str,
+    study: str,
+) -> tuple[DataFrame, DataFrame, dict[int, str]]:
+    """Extract information cg-gnn needs from SPT."""
+    extractor = FeatureMatrixExtractor(database_config_file=spt_db_config_location)
+    study_data: dict[str, Any] = extractor.extract(study=study)
+    df_cell = _create_cell_df(
+        {slide: data['dataframe'] for slide, data in study_data['feature matrices'].items()},
+        study_data['channel symbols by column name']
+    )
+    df_label, label_to_result_text = _create_label_df(
+        study_data['sample cohorts']['assignments'],
+        study_data['sample cohorts']['strata']
+    )
+    return df_cell, df_label, label_to_result_text
--- a/spatialprofilingtoolbox/cggnn/scripts/extract.py
+++ b/spatialprofilingtoolbox/cggnn/scripts/extract.py
@@ -1,0 +1,48 @@
+"""Extract information cg-gnn needs from SPT and save to file."""
+
+from argparse import ArgumentParser
+from os.path import join, exists
+from json import dump
+
+from spatialprofilingtoolbox.cggnn import extract_cggnn_data
+
+
+def parse_arguments():
+    """Process command line arguments."""
+    parser = ArgumentParser(
+        prog='spt cggnn extract',
+        description='Extract information cg-gnn needs from SPT and save to file.'
+    )
+    parser.add_argument(
+        '--spt_db_config_location',
+        type=str,
+        help='File location for SPT DB config file.',
+        required=True
+    )
+    parser.add_argument(
+        '--study',
+        type=str,
+        help='Name of the study to query data for in SPT.',
+        required=True
+    )
+    parser.add_argument(
+        '--output_location',
+        type=str,
+        help='Directory to save extracted data to.',
+        required=True
+    )
+    return parser.parse_args()
+
+
+if __name__ == "__main__":
+    args = parse_arguments()
+    df_cell, df_label, label_to_result = extract_cggnn_data(args.spt_db_config_location, args.study)
+
+    assert isinstance(args.output_location, str)
+    dict_filename = join(args.output_location, 'label_to_results.json')
+    cells_filename = join(args.output_location, 'cells.h5')
+    labels_filename = join(args.output_location, 'labels.h5')
+    if not (exists(dict_filename) and exists(cells_filename) and exists(labels_filename)):
+        df_cell.to_hdf(cells_filename, 'cells')
+        df_label.to_hdf(labels_filename, 'labels')
+        dump(label_to_result, open(dict_filename, 'w', encoding='utf-8'))
--- a/spatialprofilingtoolbox/cggnn/scripts/run.py
+++ b/spatialprofilingtoolbox/cggnn/scripts/run.py
@@ -1,16 +1,13 @@
-"Run through the entire SPT CG-GNN pipeline using a local db config."
+"""Run through the entire SPT CG-GNN pipeline using a local db config."""
+
 from argparse import ArgumentParser
 from os.path import join
 
-from pandas import DataFrame
 from pandas import read_csv
-from pandas import concat, merge
-from numpy import sort
 
-from spatialprofilingtoolbox.db.database_connection import DatabaseConnectionMaker
-from spatialprofilingtoolbox.db.database_connection import DBCredentials
+from spatialprofilingtoolbox.cggnn import extract_cggnn_data
+from spatialprofilingtoolbox.db.database_connection import DatabaseConnectionMaker, DBCredentials
 from spatialprofilingtoolbox.db.importance_score_transcriber import transcribe_importance
-from spatialprofilingtoolbox.db.feature_matrix_extractor import FeatureMatrixExtractor
 from spatialprofilingtoolbox.standalone_utilities.module_load_error import SuggestExtrasException
 try:
     from cggnn.run_all import run_with_dfs
@@ -19,7 +16,7 @@ except ModuleNotFoundError as e:
 
 
 def parse_arguments():
-    "Process command line arguments."
+    """Process command line arguments."""
     parser = ArgumentParser(
         prog='spt cggnn run',
         description='Create cell graphs from SPT tables saved locally, train a graph neural '
@@ -120,42 +117,10 @@ def parse_arguments():
     return parser.parse_args()
 
 
-def _create_cell_df(cell_dfs: dict[str, DataFrame], feature_names: dict[str, str]) -> DataFrame:
-    "Find chemical species, phenotypes, and locations and merge into a DataFrame."
-
-    for specimen, df_specimen in cell_dfs.items():
-        df_specimen.rename({ft_id: 'FT_' + ft_name for ft_id, ft_name in feature_names.items()},
-                           axis=1, inplace=True)
-        # TODO: Create phenotype columns
-        df_specimen.rename({'pixel x': 'center_x', 'pixel y': 'center_y'},
-                           axis=1, inplace=True)
-        df_specimen['specimen'] = specimen
-
-    # TODO: Reorder so that it's feature, specimen, phenotype, xy
-    # TODO: Verify histological structure ID or recreate one
-    df = concat(cell_dfs.values(), axis=0)
-    df.index.name = 'histological_structure'
-    return df
-
-
-def _create_label_df(df_assignments: DataFrame,
-                     df_strata: DataFrame) -> tuple[DataFrame, dict[int, str]]:
-    """Get slide-level results."""
-    df = merge(df_assignments, df_strata, on='stratum identifier', how='left')[
-        ['specimen', 'subject diagnosed result']].rename(
-        {'specimen': 'slide', 'subject diagnosed result': 'result'}, axis=1)
-    label_to_result = dict(enumerate(sort(df['result'].unique())))
-    return df.replace({res: i for i, res in label_to_result.items()}), label_to_result
-
-
-def retrieve_importances() -> DataFrame:
-    filename = join('out', 'importances.csv')
-    return read_csv(filename, index_col=0)
-
-
-def save_importances(_args):
-    df = retrieve_importances()
-    credentials = DBCredentials(_args.dbname, _args.host, _args.user, _args.password)
+def save_importance(dbname: str, host: str, user: str, password: str) -> None:
+    """Save cell importance scores as defined by cggnn to the database."""
+    df = read_csv(join('out', 'importances.csv'), index_col=0)
+    credentials = DBCredentials(dbname, host, user, password)
     connection = DatabaseConnectionMaker.make_connection(credentials)
     transcribe_importance(df, connection)
     connection.close()
@@ -163,20 +128,10 @@ def save_importances(_args):
 
 if __name__ == "__main__":
     args = parse_arguments()
-    extractor = FeatureMatrixExtractor(database_config_file=args.spt_db_config_location)
-    study_data: dict[str, dict] = extractor.extract(study=args.study)
-
-    df_cell = _create_cell_df(
-        {slide: data['dataframe']
-            for slide, data in study_data['feature matrices'].items()},
-        study_data['channel symbols by column name'])
-    df_label, label_to_result_text = _create_label_df(
-        study_data['sample cohorts']['assignments'],
-        study_data['sample cohorts']['strata'])
-
+    df_cell, df_label, label_to_result = extract_cggnn_data(args.spt_db_config_location, args.study)
     run_with_dfs(df_cell,
                  df_label,
-                 label_to_result_text,
+                 label_to_result,
                  args.validation_data_percent,
                  args.test_data_percent,
                  args.roi_side_length,
@@ -189,4 +144,4 @@ if __name__ == "__main__":
                  args.merge_rois,
                  args.prune_misclassified)
 
-    save_importances(args)
+    save_importance(args.dbname, args.host, args.user, args.password)
--- a/spatialprofilingtoolbox/cggnn/__init__.py
+++ b/spatialprofilingtoolbox/cggnn/__init__.py
@@ -1,2 +1,4 @@
 """Cell-graph graph neural network functionality."""
 __version__ = '0.2.1'
+
+from spatialprofilingtoolbox.cggnn.extract import extract_cggnn_data
--- a/spatialprofilingtoolbox/cggnn/scripts/run_sql.py
+++ b/spatialprofilingtoolbox/cggnn/scripts/run_sql.py
@@ -1,4 +1,5 @@
 "Run through the entire SPT CG-GNN pipeline, starting from a SPT SQL database."
+
 from argparse import ArgumentParser
 from spatialprofilingtoolbox.standalone_utilities.module_load_error import SuggestExtrasException
 try:
@@ -8,7 +9,7 @@ except ModuleNotFoundError as e:
 
 
 def parse_arguments():
-    "Process command line arguments."
+    """Process command line arguments."""
     parser = ArgumentParser(
         prog='spt cggnn run_sql',
         description='Create cell graphs from SPT SQL tables, train a graph neural network on '


spatialprofilingtoolbox/cggnn/extract.py
```diff
@@ -0,0 +1,57 @@
...
+
+from spatialprofilingtoolbox.db.feature_matrix_extractor import FeatureMatrixExtractor
+
+
+def _create_cell_df(
+    dfs_by_specimen: dict[str, DataFrame],
+    feature_names: dict[str, str],
+) -> DataFrame:
+    """Find simple and complex phenotypes, and locations and merge into a DataFrame."""
+    feature_ids_to_names = {ft_id: 'FT_' + ft_name for ft_id, ft_name in feature_names.items()}
```
CarlinLiao, line 16: How can we distinguish simple phenotypes from complex phenotypes? Inside `cg-gnn` it distinguishes between the two because the former has a `FT_` prefix and the latter have a `PH_` prefix.

spatialprofilingtoolbox/cggnn/extract.py
```diff
@@ -0,0 +1,57 @@
...
+    for specimen, df_specimen in dfs_by_specimen.items():
+        df_specimen.rename(feature_ids_to_names, axis=1, inplace=True)
+        # TODO: Create complex phenotype columns
+        df_specimen.rename({'pixel x': 'center_x', 'pixel y': 'center_y'},
+                           axis=1, inplace=True)
+        df_specimen['specimen'] = specimen
+
+    # TODO: Reorder so that it's simple phenotype, specimen, complex phenotype, x, y
+    df = concat(dfs_by_specimen.values(), axis=0)
```
CarlinLiao, line 25: We can reorder the columns here once we figure out how to separate the simple and complex phenotypes.

I don't think it actually matters in practice but this is the ordering originally defined in `cg-gnn`'s own SQL-based extraction method.

spatialprofilingtoolbox/cggnn/extract.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def _create_label_df(
+    df_assignments: DataFrame,
+    df_strata: DataFrame,
+) -> tuple[DataFrame, dict[int, str]]:
+    """Get slide-level results."""
+    df = merge(df_assignments, df_strata, on='stratum identifier', how='left')[
+        ['specimen', 'subject diagnosed result']].rename(
+        {'specimen': 'slide', 'subject diagnosed result': 'result'}, axis=1)
```
CarlinLiao, line 37: (This code is from what was already implemented in `main` in `run_sql.py`, just moved to a new file.) Is this how we want to define the labels to classify against? It appears to be a little different from what we discussed over email.

spatialprofilingtoolbox/cggnn/scripts/run.py
```diff
@@ -120,63 +117,21 @@ def parse_arguments():
...
-    return read_csv(filename, index_col=0)
-
-
-def save_importances(_args):
-    df = retrieve_importances()
-    credentials = DBCredentials(_args.dbname, _args.host, _args.user, _args.password)
+def save_importance(dbname: str, host: str, user: str, password: str) -> None:
+    """Save cell importance scores as defined by cggnn to the database."""
+    df = read_csv(join('out', 'importances.csv'), index_col=0)
+    credentials = DBCredentials(dbname, host, user, password)
```
CarlinLiao, line 123: Can this be a `DBConfig` instead of a 4-tuple input to be consistent with the run and extract methods?

spatialprofilingtoolbox/cggnn/extract.py
jimmymathews: If I recall correctly, `FeatureMatrixExtractor` doesn't provide the "composite phenotype" assignments, only the dichotomized expression matrix for the original channels of imaging ("simple phenotypes").
The reasoning was that these assignments aren't really part of the source data, they are derivative.
Still, it seems you make use of these. A little helper function would be useful to this end, perhaps: 

```py
def assign_phenotypes(expressions: DataFrame, study_name: str) -> DataFrame
```
?
This function could grab all the phenotype definitions for the study, and apply `get_mask` to each one.

CarlinLiao:
Installed this branch to test its functionality. It errored on this line after testing with "Breast cancer IMC":
```
  File "/Users/cliao/miniconda3/envs/spt_use/lib/python3.11/site-packages/spatialprofilingtoolbox/cggnn/scripts/extract.py", line 39, in <module>
    df_cell, df_label, label_to_result = extract_cggnn_data(args.spt_db_config_location, args.study)
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cliao/miniconda3/envs/spt_use/lib/python3.11/site-packages/spatialprofilingtoolbox/cggnn/extract.py", line 50, in extract_cggnn_data
    {slide: data['dataframe'] for slide, data in study_data['feature matrices'].items()},
                                                 ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
KeyError: 'feature matrices'
```
`study_data` is `extractor.extract(study=study)` so I don't see why it would break here.

spatialprofilingtoolbox/cggnn/scripts/run.py
jimmymathews: I assume you mean `DBCredentials`, but yes. Why not? Is there something I'm missing?

spatialprofilingtoolbox/cggnn/extract.py
CarlinLiao: Would this be standalone, in this file, or in `FeatureMatrixExtractor`?

jimmymathews:
Hm it seems that `FeatureMatrixExtractor` is pedantically interested in the `specimen_measurement_study` names only. This is related to the fact that it does not bother to look for composite phenotypes, since these are part of the data analysis study phase, not the data measurement phase of the study.

Still, it would much more convenient if `FeatureMatrixExtractor` took responsibility for looking up the measurement study name from the overall study handle string. In the current implemention, you have to lookup the measurement study name yourself (most likely "Breast cancer IMC - measurement"?).

CarlinLiao:
Yes. In the SQL query workflow of the standalone `cg-gnn` pip package it queries the measurement study for the targets and shape strings, the analysis study for phenotypes, and the specimen collection study for the stratification. The `FeatureMatrixExtractor` has elements sourced from all three (sub)studies, so it's interesting that they aren't wholly reflected in the current implementation.

jimmymathews:
Yes, it is a legacy aspect (like the "Feature Matrix" in the name), reflecting its original purpose of providing a familiar feature matrix representation of the portion of the dataset which is representable this way.

spatialprofilingtoolbox/cggnn/scripts/run.py
CarlinLiao: Looking back at this, this would actually error in the version committed to `main`. `args.dbname, args.host, args.user, args.password` are provided to the `save_importance` function in the script section of this file, but the `ArgumentParser` only asks for `args.spt_db_config_location` (as does `cggnn run extract`).

CarlinLiao:
The complex phenotypes features would be consistent with the feature matrix conceptualization.

spatialprofilingtoolbox/cggnn/scripts/run.py
jimmymathews: I see, I didn't catch that.

CarlinLiao: Pushing a commit to fix this...

CarlinLiao:
I wrote a test for the extract function and it errors because SPT is written in Python 3.11 but the Docker image for `set cggnn` is Python 3.7 for DGL and PyTorch compatibility. This causes the test to error on new 3.11 syntax like `match`. How should we proceed?

(It also complains about unary operators in these lines, but they're identical to other tests I assume Bash SyntaxErrors aren't breaking, and only reported if something else breaks, as the above.)
```
[ $status -eq 0 ] || echo "cggnn extract failed."

if [ $status -eq 0 ];
```

jimmymathews:
Well 3.7 is pretty old, released 2018. According to [this](https://github.com/dmlc/dgl/issues/5663), DGL now supports python 3.11. PyTorch I believe also does.
I'm trying a newer base image than `pytorch/pytorch:1.13.0-cuda11.6-cudnn8-runtime`, and it seems to be building OK.
I had to rebuild the `cg-gnn` wheel since it is built with a declared requirement of python `~3.9`.

jimmymathews:
As for the unary operator complaint, it is because your `status=...` line does not return anything, `status` is empty. I changed to:
```bash
$([ $? -gt 0 ] && [ -e "label_to_results.json" ] && [ -e "cells.h5" ] && [ -e "labels.h5" ])
status="$?"
...
```
and the error reports correctly (no unary operator complaint).

jimmymathews:
The new docker image works for me, the test extraction works. I pushed a proof of concept to branch [`cggnn_dfs3.11`](https://github.com/nadeemlab/SPT/tree/cggnn_dfs3.11). It relies on a locally-build `cg-gnn` that removes the python `~3.9` constraint.

CarlinLiao:
Extraction shouldn't run into any issues, it's the run pipeline I'm worried about. Histocartography model training and testing use esoteric features of DGL and pytorch that break in newer versions of both and Python (as of my testing last year, at least).

CarlinLiao:
Were you able to take another look at the complex phenotype extraction as well as the stratification method?

jimmymathews:
If the dependency conflicts are persistent, it probably makes sense for us to keep major functionality isolated to separate CLI tool instances, not requiring the same python runtime for SPT and for cg-gnn.

CarlinLiao:
I'm going to test if the latest changes in the histocartography repo fix the dependency issues, but in the event they don't help, yes we'll need to further isolate cggnn running, either by removing `spt cggnn run` entirely or by having it call a separate container instance that has Python 3.9 and the `cg-gnn` pip package.

jimmymathews:
What would you like me to look at for complex phenotype extraction and stratification?
As for phenotype extraction, It seems that what is needed is a classifier utility that pulls out the signatures and assesses each cell in the cell dataframe for membership in the class associated with each given signature.
That should be pretty easy. This snippet will get you started with pulling phenotypes:

```py
import json

from pandas import read_sql

from spatialprofilingtoolbox.db.database_connection import DatabaseConnectionMaker
from spatialprofilingtoolbox.db.exchange_data_formats.metrics import PhenotypeCriteria

class PhenotypeDefinitionPuller:
    @staticmethod
    def _get_phenotype(criteria_dataframe):
        positives, negatives = [
            sorted(map(str, list(criteria_dataframe[
                criteria_dataframe['coded_value']==coded_value
            ]['channel_symbol'])))
            for coded_value in (1, 0)
        ]
        return PhenotypeCriteria(positive_markers=positives, negative_markers=negatives)

    @classmethod
    def pull(cls, study):
        query = r'''
        SELECT
            cp.name as phenotype_name,
            cs.symbol as channel_symbol,
            CASE WHEN cpc.polarity='positive' THEN 1 ELSE 0 END AS coded_value
        FROM cell_phenotype_criterion cpc
        JOIN cell_phenotype cp ON cp.identifier=cpc.cell_phenotype
        JOIN chemical_species cs ON cs.identifier=cpc.marker
        JOIN study_component sc ON sc.component_study=cpc.study
        WHERE sc.primary_study='%s'
        ;
        '''
        with DatabaseConnectionMaker() as dcm:
            df = read_sql(query % study, dcm.get_connection())
        return {
            key: cls._get_phenotype(group)
            for key, group in df.groupby('phenotype_name')
        }

phenotypes = PhenotypeDefinitionPuller.pull('Breast cancer IMC')
for key, phenotype_definition in phenotypes.items():
    print('')
    print(key)
    print(phenotype_definition)

phenotypes = PhenotypeDefinitionPuller.pull('Melanoma intralesional IL2')
for key, phenotype_definition in phenotypes.items():
    print('')
    print(key)
    print(phenotype_definition)
```

If you have a local db running, you can run the above as
```sh
SINGLE_CELL_DATABASE_HOST=localhost SINGLE_CELL_DATABASE_USER=postgres SINGLE_CELL_DATABASE_PASSWORD=postgres python snippet.py
```

The output:
```txt
Luminal
positive_markers=['ESR1', 'PGR'] negative_markers=[]

Myoepithelial
positive_markers=['ACTA2', 'CK', 'EGFR', 'KRT14', 'MYC'] negative_markers=[]

Proliferative
positive_markers=['MKI67'] negative_markers=[]



Adipocyte or Langerhans cell
positive_markers=['S100B'] negative_markers=['CD14', 'CD163', 'CD20', 'CD3', 'CD4', 'CD56', 'CD68', 'CD8', 'FOXP3', 'SOX10']

B cell
positive_markers=['CD20'] negative_markers=['CD3', 'CD4', 'CD56', 'CD8', 'FOXP3', 'SOX10']

CD163+MHCII+ macrophage
positive_markers=['CD163', 'MHCII'] negative_markers=['CD20', 'CD3', 'CD56', 'CD8', 'FOXP3', 'SOX10']

CD163+MHCII- macrophage
positive_markers=['CD163'] negative_markers=['CD20', 'CD3', 'CD56', 'CD8', 'FOXP3', 'MHCII', 'SOX10']

CD4+ T cell
positive_markers=['CD3', 'CD4'] negative_markers=['CD20', 'CD56', 'CD8', 'FOXP3', 'SOX10']

CD4+ natural killer T cell
positive_markers=['CD3', 'CD4', 'CD56'] negative_markers=['CD20', 'CD8', 'FOXP3', 'SOX10']

CD4+ regulatory T cell
positive_markers=['CD3', 'CD4', 'FOXP3'] negative_markers=['CD20', 'CD56', 'CD8', 'SOX10']

CD4+/CD8+ T cell
positive_markers=['CD3', 'CD4', 'CD8'] negative_markers=['CD20', 'CD56', 'FOXP3', 'SOX10']

CD68+MHCII+ macrophage
positive_markers=['CD68', 'MHCII'] negative_markers=['CD20', 'CD3', 'CD56', 'CD8', 'FOXP3', 'SOX10']

CD68+MHCII- macrophage
positive_markers=['CD68'] negative_markers=['CD20', 'CD3', 'CD56', 'CD8', 'FOXP3', 'MHCII', 'SOX10']

CD8+ T cell
positive_markers=['CD3', 'CD8'] negative_markers=['CD20', 'CD4', 'CD56', 'FOXP3', 'SOX10']

CD8+ natural killer T cell
positive_markers=['CD3', 'CD56', 'CD8'] negative_markers=['CD20', 'CD4', 'FOXP3', 'SOX10']

CD8+ regulatory T cell
positive_markers=['CD3', 'CD8', 'FOXP3'] negative_markers=['CD20', 'CD4', 'CD56', 'SOX10']

Double negative regulatory T cell
positive_markers=['CD3', 'FOXP3'] negative_markers=['CD20', 'CD4', 'CD56', 'CD8', 'SOX10']

Natural killer T cell
positive_markers=['CD3', 'CD56'] negative_markers=['CD20', 'CD4', 'CD8', 'FOXP3', 'SOX10']

Natural killer cell
positive_markers=['CD56'] negative_markers=['CD20', 'CD3', 'CD4', 'CD8', 'FOXP3', 'S100B', 'SOX10']

Nerve
positive_markers=['CD56', 'S100B'] negative_markers=['CD14', 'CD163', 'CD20', 'CD3', 'CD4', 'CD68', 'CD8', 'FOXP3', 'SOX10']

Other macrophage/monocyte CD14+
positive_markers=['CD14'] negative_markers=['CD163', 'CD20', 'CD3', 'CD56', 'CD68', 'CD8', 'FOXP3', 'SOX10']

Other macrophage/monocyte CD4+
positive_markers=['CD4'] negative_markers=['CD163', 'CD20', 'CD3', 'CD56', 'CD68', 'CD8', 'FOXP3', 'SOX10']

T cell/null phenotype
positive_markers=['CD3'] negative_markers=['CD20', 'CD4', 'CD56', 'CD8', 'FOXP3', 'SOX10']

Tumor
positive_markers=['SOX10'] negative_markers=['CD20', 'CD3', 'CD4', 'CD8', 'FOXP3']
```

spatialprofilingtoolbox/cggnn/extract.py
CarlinLiao: @jimmymathews Regarding stratification, this implementation uses only "subject diagnosed result" but not "stratum_identifier", "local temporal position indicator", or "subject diagnosed condition".

What do we want the model `spt cggnn extract` supports to do? In prior iterations, it's been to predict, on pre-treatment slides only, if the patient the tissue sample is from will respond to treatment. Do we want to continue this, or generalize?

CarlinLiao:
I've yet to setup a workflow that gives me a local db to query. Perhaps a tutorial in `docs/` would be good for that.

Is the plan here to add in functionality to `spt cggnn extract` so that it generates complex phenotype features itself, or to extend `FeatureMatrixExtractor` so that it extracts complex phenotypes in addition to simple ones?

jimmymathews:
I think the functionality is appropriate for a new class or functions near `FeatureMatrixExtractor`.
If you want deep integration with `FeatureMatrixExtractor`, such that the cell dataframe has new columns for each phenotype, you'll have to make some decision for naming these new columns. You could just use the long names as in the snippet, but as you have pointed out above this hides the distinction between the original-channel-associated columns and the phenotype columns. (Note that "simple" and "complex" phenotype is personal terminology of my own that will be confusing to most others. "Phenotype" refers to the latter concept normally.)
I would probably keep the phenotype columns in a separate dataframe to avoid that problem.

CarlinLiao:
If the phenotypes are kept in a separate DataFrame, then `FeatureMatrixExtractor` would just return another DataFrame in its result dictionary, right? That's simple to understand. `spt cggnn extract` can handle merging the chemical species(?) and phenotypes together in its own logic.

Would you like me to try implementing this, or would you rather do it yourself since you're more familiar with the architecture of the extractor?

spatialprofilingtoolbox/cggnn/extract.py
jimmymathews: Prediction of treatment response from pre-treatment slides is certainly not the only study design that we support.

Although one cannot, in advance, write a single description of the outcome variable that will always be of greatest interest in any study, the `SampleStratificationCreator` does a pretty good job of automatically creating a best guess at a reasonable granularity, given the facts about the samples available according to the schema.

The result of `SampleStratificationCreator`'s work is the outcome variable that I endorse for the use case of general-purpose downstream analysis.

It is defined by the fields named `stratum_identifier` (an integer) and `sample`, in a supplementary table called `sample_strata`.

In this use case, one shouldn't be reading `local_temporal_position_indicator` or `subject_diagnosed_condition` or `subject_diagnosed_result` at all. In this use case, these values are just metadata tags for the strata.

jimmymathews:
Sure, please give it a try. I don't think it requires too much architectural knowledge.
And yeah I was imagining another dataframe in the dict, beside the existing one.

(I'm thinking of the primary columns as "channels" as in "imaging channels". The "chemical species" terminology is the pedantic description of the target indicated by the imaging modality. The column name is the protein's gene name, so yeah this is the name of the "chemical species" indicated, but again no one else is going to call this the "chemical species".)

spatialprofilingtoolbox/cggnn/extract.py
CarlinLiao: (`sample` here means the tissue sample, right? I don't see why the `sample` is relevant to the stratification aside from samples being the granularity at which we're stratifying.)

As I envision it, the `spt cggnn extract` command isn't for "general-purpose downstream analysis" but to extract the specific information the user wants to train a model on. (If this isn't the case, we should write a separate command that does so, as I don't think this should be within the scope of the standalone `cg-gnn` package.)

So, if prediction of treatment response isn't the only study design we want to support, it sounds like what we should do is provide the person calling `spt cggnn extract` some way to identify which strata they want to use. For example, in the treatment response use case, they should be given the option of only selecting strata from before treatment. In most use cases, I don't think the user would want every stratum possible.

CarlinLiao:
I'm not sure we're aligned on what a sufficient amount of architectural knowledge is, but I'll give it a crack.

spatialprofilingtoolbox/cggnn/extract.py
jimmymathews: I see, I misunderstood your earlier remark that you want "to not have to study the intricacies of the database to run the cg-gnn pipeline". I understood it to mean that you want labels like `1`, `2`, and `3`, and you don't want to be involved with coming up with finer outcome data using the actual study data.

But it seems that you meant that you want *at runtime of this extract tool* to not have to study the intricacies of the database, but that you are willing to study them when writing the extract tool source code.

In this case, the convenience sample stratifier isn't for you. Just use direct SQL queries to `diagnosis`, `intervention`, `subject`, `specimen_collection_process`, etc. You could probably make a nice little CLI dialog that solicits from the user on the basis of what's already known to be available.

jimmymathews: The whole design constraint for `SampleStratificationCreator` was automation with no user intervention, but that's the opposite of what you're describing.

CarlinLiao: Would this be built into the `FeatureMatrixExtractor` like the convenience stratifier is, or handled seperately?

And to do that we'd have to guess at all the different ways a user would want to slice up and stratify the samples... or just give them every option in the database. At which point, the user may as well write their own SQL queries.

Sure we can't just support only treatment response for now, and leave the implementation of alternative classifications as ~an exercise for the reader~ future work?

jimmymathews: I don't think any of our other pre-curated studies have 2 cohorts that are both pre-treatment and distinguished by response.

jimmymathews: Maybe 1 other?

CarlinLiao: So we'd have to think of at least one classification method for each study, and they're all different.

build/cggnn/Dockerfile
```diff
@@ -1,18 +1,28 @@
-FROM pytorch/pytorch:1.13.0-cuda11.6-cudnn8-runtime
+FROM pytorch/pytorch:2.0.1-cuda11.7-cudnn8-runtime
```
CarlinLiao, line 1: I've updated `cg-gnn` and in the original repo it targets cuda 11.8, but there doesn't appear to be a Docker image for this version yet.

jimmymathews:
For the phenotype definition puller, note that this is substantially implemented already [here](https://github.com/nadeemlab/SPT/blob/9a57ec8d205e5e9561dcad4c55c2284c95d16854/spatialprofilingtoolbox/db/phenotypes.py#L50):

https://github.com/nadeemlab/SPT/blob/9a57ec8d205e5e9561dcad4c55c2284c95d16854/spatialprofilingtoolbox/db/phenotypes.py#L50

Usage example from the [`querying.py`](https://github.com/nadeemlab/SPT/blob/main/spatialprofilingtoolbox/db/querying.py) module:
```py
PhenotypesAccess(cursor).get_phenotype_criteria(study, phenotype_symbol)
```

jimmymathews:
Tests now pass.
Is this mergeable? You could make an issue out of the planned work related to outcome-variable provision, so as not to block merging of this PR (which is just about the major extractions from the db).

CarlinLiao:
Confirmed that the four additional columns are due to the four phenotypes picked up by the accessor. (Whether or not 4 phenotypes is accurate is another question.)

Something I'm noticing in the logs, the channels/features are _floats_ and not binary or true-false values, which I assumed they were when doing the positive and negative matching against the phenotype signatures. Consequently, every phenotype for ever cell is reporting as a 1. What should be done instead?

CarlinLiao:
Why is the test for `db/FeatureMatrixExtractor` in `test/workflow`?

CarlinLiao:
Perhaps we should replace references to chemical species / channels as "features" in order to disambiguate between the features being extracted (channels, phenotypes, strata, etc.) and 

This includes replacing the `f{'F{I}'}` prefix with something else, like `f{'C{I}'}`.

jimmymathews:
- Depending on which test, a different test dataset might be used. The number of phenotypes in the test datasets is either 4 or 3. The 2 mel test datasets both have 4 phenotypes, that's probably right.
- Regarding the data type in the dataframes produced by FeatureMatrixExtractor: Float values of 0 or 1 seem ok, as well as True vs False booleans, or ints. All can faithfully represent the indication, with boolean probably being the best choice for the purpose of bitwise operations and storage. We just need the phenotype assignment function to work on whatever we choose.
-  I think the only reason for the FeatureMatrixExtractor test to be in the workflow area vs db is that the preloaded data used for the workflow testing is much slimmer and more trivial than in the db testing. Here it shows that these are application/integration tests and not library unit tests.
- Between the terms "chemical species" and "channel" in source code we should prefer "channel". That change from e.g. F1 to C1 is fine. Keep in mind that the purpose of using column names like "F2", "F3", ... was for them to be arbitrary -- you're meant to look up the actual names in the supplementary dictionary when you need the name. So "F2" and "F3" may as well be randomly assigned uniquely identifying strings. If we can find some kind of promise in pandas docs that any type of string is allowed as a column name, we could deprecate these altogether and directly put the channel names and phenotype names here.

jimmymathews:
Tests now pass for me.

CarlinLiao:
What's the purpose of `FeatureMatrixExtractor.redact_dataframes()`? @jimmymathews

jimmymathews:
> What's the purpose of `FeatureMatrixExtractor.redact_dataframes()`? @jimmymathews

It makes it possible to create a JSON dump of everything else, for inspection.

CarlinLiao:
Ok, I've changed some functionality that arise from each other.
1. I've integrated channel and phenotype names into the cell extract dataframes,
2. This slimmed down the extract return packet from 4 entries to just two, cells and stratification, so
3. I split the `extract` function's concern to focus solely on returning cell information, as often the stratification return result was unused, such as in the case of a single-specimen extraction, so that's unnecessary work. Stratification information is now a separate function, `extract_cohorts`
4. Focusing `extract` gave me the opportunity to be more precise about the return type, so now it returns a dict mapping specimen string names to `MatrixBundle`s, which are a `dataclass` with three properties: `dataframe`, `filename`, and `continuous_dataframe` (optional)

I think this is a really good change for the sake of clarity, but I may not have fixed the tests correctly. I think I caught all downstream changes from the return type, at least.

Next on the docket is adding in the proper stratification interpretation to `cggnn extract`, AKA the original purpose of this PR before I scope creeped it with all these modifications to `FeatureMatrixExtractor`.

jimmymathews:
I think dataclasses are fine, the only FastAPI thing is that FastAPI is extra helpful when the response types are hinted with pydantic types. And query parameters I guess also.

CarlinLiao:
Ok, I don't think FME and FastAPI touch at all either.

CarlinLiao:
Resolved picking out strata to classify on by adding some fake interactivity by having the cggnn module print out the entire `strata` output from FME for the user to observe. From that, they can (manually) pick the strata they want to use by their identifier and then pass them to the extract command (by typing them one by one after the argument flag).

CarlinLiao:
All tests pass for me. Are we ready to merge?

jimmymathews:
I'll do tests and review now.

spatialprofilingtoolbox/cggnn/extract.py
```diff
@@ -0,0 +1,72 @@
...
+def _create_label_df(
+    df_assignments: DataFrame,
+    df_strata: DataFrame,
+    strata_to_use: list[int] | None,
+) -> tuple[DataFrame, dict[int, str]]:
+    """Get slide-level results."""
+    df_assignments = df_assignments.set_index('specimen')
+    df_strata = df_strata.set_index('stratum identifier')
+
+    # Filter for strata to use
```
jimmymathews, line 32: I would prefer that we use functions `filter_for_strata_to_use`, `drop_same_columns`, `compress_remaining_columns_into_str`, etc., instead of comment-headed blocks. This would break down the function into smaller units but also reduce machine-unaccountable comments.
This is not essential however and shouldn't block merge.

spatialprofilingtoolbox/cggnn/scripts/extract.py
```diff
@@ -0,0 +1,63 @@
...
+    parser.add_argument(
+        '--spt_db_config_location',
+        type=str,
+        help='Location of the SPT DB config file.',
+        required=True
+    )
+    parser.add_argument(
+        '--study',
+        type=str,
+        help='Location of the SPT DB config file to use.',
```
jimmymathews, line 25: Copy and paste typo.

jimmymathews:
Test `test_autocomputed_squdpy.py` fails. I think what is going on is that the feature identifiers minted for these "autocomputed squidpy metrics" are not consistent with the ones that were minted before the changes, for whatever reason. This could happen if new metrics are computed beforehand, using up the identifier namespace, or if some are no longer computed beforehand.

We could relax the test and just ensure that the set of feature vectors (identifiers omitted) is still the same. That would mean making changes [here](https://github.com/nadeemlab/SPT/blob/06a0f8528b9b82bf6d4ba42d7df477fe6c63d8de/test/db/module_tests/test_autocomputed_squidpy.py#L28) to omit identifiers from the expected row set and the actual row set before testing for equality.

jimmymathews:
I'm also getting test failure for `test_proximity_pipeline.sh`:
```
  File "/usr/local/lib/python3.11/dist-packages/spatialprofilingtoolbox/workflow/scripts/aggregate_core_results.py", line 24, in <module>
    integrator.calculate(**args)
  File "/usr/local/lib/python3.11/dist-packages/spatialprofilingtoolbox/workflow/phenotype_proximity/integrator.py", line 36, in calculate
    self.export_feature_values(core_computation_results_files, data_analysis_study)
  File "/usr/local/lib/python3.11/dist-packages/spatialprofilingtoolbox/workflow/phenotype_proximity/integrator.py", line 66, in export_feature_values
    self.send_features_to_uploader(feature_uploader, core_computation_results_files)
  File "/usr/local/lib/python3.11/dist-packages/spatialprofilingtoolbox/workflow/phenotype_proximity/integrator.py", line 72, in send_features_to_uploader
    stage_proximity_feature_values(
  File "/usr/local/lib/python3.11/dist-packages/spatialprofilingtoolbox/workflow/common/proximity.py", line 70, in stage_proximity_feature_values
    _phenotype_identifier_lookup(row['Phenotype 1'], channel_symbols_by_column_name),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/dist-packages/spatialprofilingtoolbox/workflow/common/proximity.py", line 59, in _phenotype_identifier_lookup
    raise ValueError(f'Did not understand meaning of specifier: {handle}')
ValueError: Did not understand meaning of specifier: C B2M
```

It seems that the "API" break for FeatureMatrixExtractor (specifically the column-to-name lookup) isn't accounted for.

CarlinLiao:
`test_proximity_pipeline` should was as simple as fixing some regex, but now it's failing at the same feature specification area as `test_autocomputed_squidpy`. Hopefully this'll make it easier to debug for me since any squidpy tests will autofail on my machine because squidpy can't install on my MacBook.

spatialprofilingtoolbox/db/squidpy_metrics.py
```diff
@@ -42,39 +36,31 @@ def create_and_transcribe_squidpy_features(
...
-            continue
-        feature_uploader.stage_feature_value((symbol,), sample, value)
+    for column in df.columns:
+        if column.startswith('C '):
+            symbol = column[2:]
+            criteria = PhenotypeCriteria(positive_markers=[symbol], negative_markers=[])
+            value = compute_squidpy_metric_for_one_sample(df, [criteria], 'spatial autocorrelation')
+            if value is None:
+                continue
+            feature_uploader.stage_feature_value((symbol,), sample, value)
```
CarlinLiao, line 55: Both old and new versions of this function create and transcribe spatial autocorrelation for only channels (simple phenotypes). Do we want to do autocorrelation for (complex) phenotypes as well?

(It might require some another input parameter since the columns will have the complex phenotype name, but not its signature, so that'll have to be fetched from elsewhere.)

CarlinLiao:
The latest commit should fix the inconsistently minted feature specifications without relaxing the tests. (The tests that can run on my M1 Mac pass, but that leaves out a lot of tests.)

CarlinLiao:
Almost all tests pass now and merging this PR will quash several open issues. The only exception are the proximity and squidpy module tests in apiserver, which run but result in different values compared to expected. Not totally sure why, but this could be a case of updated computations leading to new values rather than the output being wrong.

jimmymathews:
The tests that check computed feature values (squidpy and proximity) are failing for the first time at the commit that introduces a new way to compute masks, see these lines:

https://github.com/nadeemlab/SPT/blob/f3f6178eef9fb9f321d615b630806e450bbe5a84/spatialprofilingtoolbox/workflow/common/squidpy.py#L39-L47

Most likely these masks differ. I'll try to find out which one is correct, and if the new one is not correct we can correct it.

jimmymathews:
The masks in this version were wrong according to one example I checked, and the previously computed masks were correct. I think this was due to the `.all` pandas function being applied to 0 and 1 values. I converted to `bool` in the previous commit, and the mask in the example I checked is now correct.

spatialprofilingtoolbox/db/squidpy_metrics.py
jimmymathews: I don't think it's necessary to do this, since the primary use case is the ondemand anyway.

jimmymathews:
The squidpy metrics test now passes.
Working on proximity metrics test.

jimmymathews:
All tests pass!

spatialprofilingtoolbox/db/feature_matrix_extractor.py
```diff
@@ -96,163 +131,159 @@ def _extract(self,
...
     @staticmethod
-    def _create_feature_matrix_row(centroid, binary, number_channels):
+    def _create_feature_matrix_row(
+        centroid: tuple[float, float],
+        binary: list[str],
+        number_channels: int,
+    ) -> list[float | int]:
         template = '{0:0%sb}' % number_channels   # pylint: disable=consider-using-f-string
-        feature_vector = [int(value) for value in list(template.format(binary)[::-1])]
+        feature_vector: list[int] = [int(value) for value in list(template.format(binary)[::-1])]
```
CarlinLiao, line 228: This is the line we would change if we wanted to turn this from ints to bools (not including all the downstream changes we'd need to do...)

jimmymathews (approved):
I did a pass through each file, looks OK to me. Now that tests pass it is ready to merge.

PR #198, opened 2023-08-24
Added security headers library for fastApi

wooferclaw:
Security fix for HSTS issue. Not yet tested.


jimmymathews:
It looks ok, but let's add at least 1 test that the `secure` library is doing what we expect.
This can be a test in e.g. `test/apiserver/unit_tests/` that does a handful of requests and inspects the headers. We can "pin" the expected headers with assertions of exact contents, then we will possibly know if the `secure` library is changing anything on us in the future or is injecting some vulnerability.

jimmymathews:
I added a minimal test of the headers for one simple endpoint. The test just checks that the response headers are equal to a certain list. If this changes in future test runs we will know that the secure library is acting differently.

The headers as reported by curl:
```txt
< HTTP/1.1 200 OK
< server: uvicorn
< content-length: 97
< content-type: application/json
< strict-transport-security: max-age=63072000; includeSubdomains
< x-frame-options: SAMEORIGIN
< x-xss-protection: 0
< x-content-type-options: nosniff
< referrer-policy: no-referrer, strict-origin-when-cross-origin
< cache-control: no-store
< 
{ [97 bytes data]
* Connection #0 to host spt-apiserver-testing left intact
[{"handle":"Melanoma intralesional IL2","display_name_detail":"Cancer Immunology Research 2022"}]
```

jimmymathews:
All test pass so I am merging.

Issue #202, opened 2023-08-25
Enable specialization of on-demand service for one of the 3 computation types


jimmymathews:
Update the ondemand service to optionally allow selection of just 1 of the 3 providers (squidpy, counts, proximity), selectable at the CLI `start` command.

This way in deployment, we can use 3 separate containers to handle the different types of requests, allow simultaneous/parallel initialization, etc. Related to [spt-dashboard 48](https://github.com/nadeemlab/spt-dashboard/issues/48).

PR #203, opened 2023-09-07
allow starting only one ondemand service

CarlinLiao:
Pursuant to #202. All tests pass. This should be ready for @wooferclaw to build on.


jimmymathews:
For this we will also need to update `ondemand/service_client.py`:

https://github.com/nadeemlab/SPT/blob/93a8be3dcaf389acda6108746a9138cbb117cd32/spatialprofilingtoolbox/ondemand/service_client.py#L156-L159

Some "service" parameter is needed that should trigger the client to look in a service-specific environment variable for the hostname.

Also the calls to `OnDemandRequester` should be possible to service-specialize.

jimmymathews:
I have tested this locally in minimal setup. The `--service=...` flag does seem to cause only the specified service to start up. And the `OnDemandRequester` client works with `service='...'`.

PR #204, opened 2023-09-12
Upload importance scores via cggnn command

CarlinLiao:
This adds a command to `cggnn` that will upload a CSV of importance scores to a provided database config, and an API server endpoint to return importance scores for a given study.

It also reorganizes `*Puller` classes into a subdirectory in a way that doesn't change functionality, among other things.


spatialprofilingtoolbox/db/exchange_data_formats/metrics.py
```diff
@@ -85,3 +85,9 @@ class UMAPChannel(BaseModel):
     """
     channel: str
     base64_png: str
+
+
+class CGGNNImportanceRank(BaseModel):
+    """The importance ranking of histological structures in a study."""
+    histological_structure_id: int
```
jimmymathews, line 92: A warning: In the SQL schema definition, the identifiers are generally strings (`VARCHAR(512)`). In the current software implementation, `histological_structure`'s identifier field is always filled with a string representation of an integer, but that isn't guaranteed. And the sort order differs from the int sort order accordingly. (In  `SELECT ... ORDER BY ...` or in the Python functions where these IDs are still lying around as strings).

jimmymathews:
All tests pass, ready to merge.

jimmymathews:
Doing merge gymnastics.

jimmymathews:
Tests pass after merge-in.

Issue #232, opened 2023-10-31
Supporting `cg-gnn` version 0.2


CarlinLiao:
I've made a significant upgrade to the `cg-gnn` pip package with four major changes:
* Save graphs when asked as a single `graphs.bin`, accompanied by a single `graph_info.pkl` metadata file and a `feature_names.txt` file that gives interpretable names to the graph data columns, for nicer interplay with Nextflow.
* Makes the way graphs are passed between functions internally consistent.
* Support fixing a random seed for reproducibility.
* Allow graph creation from unlabeled specimens.

The last two changes will impact SPT directly. Supporting the random seed functionality should be as simple as adding another CLI input arg and changing the workflow test to exploit the reproducibility, assuming it works as intended, but the latter may be more involved.

The purpose of a machine learning model is to predict unlabeled data, but, as implemented, when using `spt cggnn extract` we're only fetching specimen with locked-down strata. For example, the strata of the urolithelial dataset is as follows
```
   stratum identifier local temporal position indicator                      subject diagnosed condition subject diagnosed result
0                  14               Before intervention  Response to immune checkpoint inhibitor therapy                Responder
1                  15               Before intervention  Response to immune checkpoint inhibitor therapy            Non-responder
15                 16               Before intervention  Response to immune checkpoint inhibitor therapy                         
```
Right now when I extract data I simply drop strata 16, but if we could add some functionality so that strata 16 is extracted as "unlabeled" data, that could be one way to naturally enable this functionality. The other study with unknown strata is melanoma CyTOF, which looks slightly different.
```
   stratum identifier local temporal position indicator subject diagnosed condition subject diagnosed result
0                   9               Before intervention         Response to therapy                Responder
1                  10               Before intervention         Response to therapy            Non-responder
30                 11                                                                                       
```
Perhaps we identify strata that have more empty values that other strata as "unlabeled"?

I'd consider this issue closed when SPT is updated to use `cg-gnn` version 0.2.1, which will include all these changes.

sanadeem:
I will assume you are using https://pytorch.org/docs/stable/notes/randomness.html. This normally leads to drastic drop in performance. As you suggested, finding consistent cells across multiple, fresh runs is the best way to proceed.

jimmymathews:
The strict reproducibility is mainly for the testing setup. In the real runs the model performance has more priority than strict reproduction.

[...]

PR #233, opened 2023-11-01
Update to support cg-gnn 0.2.1

CarlinLiao:
Pursuant to #232.

This is potentially ready to merge, all tests pass on my machine*, but I'm going to wait to turn this into a full PR until I can experiment with the extract functionality on the live scstudies database.

\* This is actually untrue. The first test or two in db and ondemand unit tests, and well as db module tests, fail, but I don't think this is attributable to the change made. Investigation reveals that the errors are caused by my docker environment not finishing setup in the 15 second wait period before test blocks begin running. Unless my changes have slowed down db setup dramatically, which nothing in the changes suggest it should, this is an issue with my hardware being too slow for the test suite rather than any code changes.


jimmymathews:
If the timeout is too severe for your hardware, you can increase it in your specific setup.

CarlinLiao:
How do I up the timeout?

CarlinLiao:
I believe I've fixed the env issue where Torch kept wanting to rebuild itself. I've moved onto a string interpolation issue that has me losing my mind. It's not related to the `cg-gnn` pip package version, but rather something uncaught because I didn't specify a `target_name` in the test because the test dataset so small that if I did it'd break for other reasons.

The target name is the name of a study phenotype or channel. The problem is due to most of our target names having a space in them, e.g., "P Melanoma". Somewhere along the string interpretation path between Nextflow, bash, and a bash subcommand, the quotation marks get garbled, causing an error. I've tried dozens of combinations of single quotes, double quotes, escaped quotes, equal signs, etc, but they all error because either what's being echoed is `--target_name P Melanoma` or, even more mystifyingly, `--target_name '"P' 'Melanoma"'`. Somehow, the subcommand is being correctly interpreted, but after coming out of the subcommand, either Nextflow is going in and doing some more weird substitutions or I don't understand bash substitution as well as I think I do.

Here's an example:
```nextflow
    shell:
    '''
    set -x
    spt cggnn run \
        --spt_db_config_location "!{db_config_file}" \
        --study "!{study_name}" \
        $(if [[ "!{strata}" != all ]]; then echo "--strata !{strata}"; fi) \
        --validation_data_percent !{validation_data_percent} \
        --test_data_percent !{test_data_percent} \
        $(if [[ "!{disable_channels}" = true ]]; then echo "--disable_channels"; fi) \
        $(if [[ "!{disable_phenotypes}" = true ]]; then echo "--disable_phenotypes"; fi) \
        --cells_per_slide_target !{cells_per_slide_target} \
        $(if [[ "!{target_name}" != none ]]; then echo "--target_name"; fi) \
        $(if [[ "!{target_name}" != none ]]; then echo \\"!{target_name}\\"; fi) \
        $(if [[ "!{in_ram}" = true ]]; then echo "--in_ram"; fi) \
        --batch_size !{batch_size} \
        --epochs !{epochs} \
        --learning_rate !{learning_rate} \
        --k_folds !{k_folds} \
        --explainer_model "!{explainer_model}" \
        $(if [[ "!{merge_rois}" = true ]]; then echo "--merge_rois"; fi) \
        $(if [[ "!{prune_misclassified}" = true ]]; then echo "--prune_misclassified"; fi) \
        --output_prefix "!{output_prefix}" \
        $(if [[ "!{upload_importances}" = true ]]; then echo "--upload_importances"; fi)
    '''
```

Output:
```sh
Command error:
  ++ [[ 1 2 != all ]]
  ++ echo '--strata 1 2'
  ++ [[ false = true ]]
  ++ [[ false = true ]]
  ++ [[ P Melanoma != none ]]
  ++ echo --target_name
  ++ [[ P Melanoma != none ]]
  ++ echo '"P' 'Melanoma"'
  ++ [[ true = true ]]
  ++ echo --in_ram
  ++ [[ true = true ]]
  ++ echo --merge_rois
  ++ [[ false = true ]]
  ++ [[ false = true ]]
  + spt cggnn run --spt_db_config_location spt_db.config --study 'Melanoma CyTOF ICI' --strata 1 2 --validation_data_percent 15 --test_data_percent 0 --cells_per_slide_target 5000 --target_name '"P' 'Melanoma"' --in_ram --batch_size 1 --epochs 5 --learning_rate 1e-3 --k_folds 0 --explainer_model pp --merge_rois --output_prefix mci
  usage: spt cggnn run [-h] --spt_db_config_location SPT_DB_CONFIG_LOCATION --study STUDY [--strata STRATA [STRATA ...]]
                       [--validation_data_percent VALIDATION_DATA_PERCENT] [--test_data_percent TEST_DATA_PERCENT] [--disable_channels]
                       [--disable_phenotypes] [--roi_side_length ROI_SIDE_LENGTH] [--cells_per_slide_target CELLS_PER_SLIDE_TARGET]
                       [--target_name TARGET_NAME] [--in_ram] [-b BATCH_SIZE] [--epochs EPOCHS] [-l LEARNING_RATE] [-k K_FOLDS]
                       [--explainer_model EXPLAINER_MODEL] [--merge_rois] [--prune_misclassified] [--output_prefix OUTPUT_PREFIX]
                       [--upload_importances]
  spt cggnn run: error: unrecognized arguments: Melanoma"
```

Maybe building up the command as a string and then using `eval` would work, but I don't understand why that'd be any different than calling it directly.

jimmymathews:
That sounds frustrating.

First of all, the exclamation point syntax is not normal for bash. From the [documentation](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion):

*The basic form of parameter expansion is ${parameter}. The value of parameter is substituted. The parameter is a shell parameter as described above (see [Shell Parameters](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameters)) or an array reference (see [Arrays](https://www.gnu.org/software/bash/manual/bash.html#Arrays)). The braces are required when parameter is a positional parameter with more than one digit, or when parameter is followed by a character that is not to be interpreted as part of its name.*

*If the first character of parameter is an exclamation point (!), and parameter is not a nameref, it introduces a level of indirection. Bash uses the value formed by expanding the rest of parameter as the new parameter; this is then expanded...*

I think "parameter" is bash's name for variables.

jimmymathews:
I also just noticed that your exclamation points are *before* the braces ... I've never seen this syntax in bash. Possibly you're accidentally invoking [history expansion](https://www.gnu.org/software/bash/manual/bash.html#Event-Designators)?

jimmymathews:
It seems you're using Nextflows ["`shell` block" feature](https://www.nextflow.io/docs/latest/process.html#shell).

That feature is described in the docs as an alternative to the `script` block, but the `shell` block seems to have a lot of caveats regarding precisely this issue (string delimitation and variable substitution). I would suggest switching back to `script`.

CarlinLiao:
I originally switched from `script` to `shell` because I was using so many subshell commands that making a clear distinction between `$` for Bash  and `!` for Nextflow was very useful to me. I've tried switching back but I haven't yet found it to able to resolve the issue.

Stripping down the test example to just the problematic flag,
```
    script:
    """
    set -x
    spt cggnn run \
        \$(if [[ "$target_name" != none ]]; then echo "--target_name \\"$target_name\\"" ; fi)
    """
```
results in
```
Command executed:

  set -x
  spt cggnn run         $(if [[ "P Melanoma" != none ]]; then echo "--target_name \"P Melanoma\"" ; fi)
```
which gives
```
Command error:
  ++ [[ P Melanoma != none ]]
  ++ echo '--target_name "P Melanoma"'
  + spt cggnn run --target_name '"P' 'Melanoma"'
  usage: spt cggnn run [-h] ...
```
Reducing the backslashes in the nf file from pairs to individuals was equivalent to including no backslashes at all. This gives this output:
```
spt cggnn run         $(if [[ "P Melanoma" != none ]]; then echo "--target_name "P Melanoma"" ; fi)
```
and
```
  ++ [[ P Melanoma != none ]]
  ++ echo '--target_name P' Melanoma
  + spt cggnn run --target_name P Melanoma
```

Given these experiments, my guess is that it's a Bash thing and not a Nextflow thing. For some reason ```spt cggnn run         $(if [[ "P Melanoma" != none ]]; then echo "--target_name \"P Melanoma\"" ; fi)``` is getting garbled into ```spt cggnn run --target_name '"P' 'Melanoma"'``` when the command is actually called. I can confirm this behavior outside of nextflow with the command
```
set -x; spt cggnn run         $(if [[ "P Melanoma" != none ]]; then echo "--target_name \"P Melanoma\"" ; fi)
```

Given that you have more experience writing Bash scripts than I do, how would you write a script from scratch that does what I'd like to do for these flags?

CarlinLiao:
Building the run command as a string and then `eval`ing appears to be working.

CarlinLiao:
Test pass. At least, as many tests as can pass on my hardware until I figure out how to extend the wait for servers to come up.

jimmymathews:
I made a couple updates:
1.  Reverted to a bash script block for consistency.
2. For a little more maintainability, broke up the eval'ed command string back into a normal command, with a couple pre-prepared option strings saved to separate variables.
3. Use escaping to set apart the bash variables and subshells for the benefit of nextflow's parsing.
4. Make sure that the tests' exit code is forwarded up the chain to cause an actual error in case of failure.

jimmymathews:
To change the container startup timeout, look for `container_wait_time` in `build_scripts`.

spatialprofilingtoolbox/workflow/assets/cggnn.nf
```diff
@@ -85,29 +88,40 @@ process run_cggnn {
...
+    #!/bin/bash
+
+    strata_option=\$( if [[ "${strata}" != "all" ]]; then echo "--strata ${strata}"; fi)
+    disable_channels_option=\$( if [[ "${disable_channels}" == "true" ]]; then echo "--disable_channels "; fi)
+    disable_phenotypes_option=\$( if [[ "${disable_phenotypes}" = "true" ]]; then echo "--disable_phenotypes"; fi)
+    target_name_option=\$( if [[ "${target_name}" != "none" ]]; then echo "--target_name '${target_name}'"; fi)
+    in_ram_option=\$( if [[ "${in_ram}" == "true" ]]; then echo "--in_ram"; fi)
+    merge_rois_option=\$( if [[ "${merge_rois}" == "true" ]]; then echo "--merge_rois"; fi)
+    prune_misclassified_option=\$( if [[ "${prune_misclassified}" == "true" ]]; then echo "--prune_misclassified"; fi)
+    upload_importances_option=\$( if [[ "${upload_importances}" == "true" ]]; then echo "--upload_importances"; fi)
```
CarlinLiao, line 102: Do these strings no longer need spaces after them? If so, we should remove the space after `--disable_channels` for consistency.

jimmymathews: That is correct.
Typically in bash stuff, if whitespace is needed, it should be added at the last moment, since many intermediate steps are whitespace killers.

CarlinLiao:
Ready to merge?

jimmymathews:
Ready.

PR #238, opened 2023-11-03
handle json-incompliant floats from OnDemandProvider

### LINKED ISSUE #218: Null data values conversion to JSON
Occasionally the co-occurrence metric (and perhaps others) cannot be computed due to sparsity etc., in which case a `nan` value is sometimes returned. This should be `None` instead to support automatic conversion to JSON for return value to client applications.

```txt
ValueError: Out of range float values are not JSON compliant
```

CarlinLiao:
To close #218, I edited PendingProvider to convert `inf` and `nan` values fetched from the quantitative feature value table to `None` to address downstream issues with conversion to JSON. (Tests pass.)

If we don't want to convert `inf` and `nan` values until right before the returned values are converted to JSON, instead we could apply a similar processing step to the returned `metrics` dict in `OnDemandRequestHandler._handle_squidpy_request` and `_handle_proximity_request`.


jimmymathews:
Haven't completely 100% tested in application, but the tests pass and the resulting containers run without errors so far. Should still try to reproduce the nan's and test in final usage.

Issue #226, opened 2023-10-17
Slim down expression quantification index


jimmymathews:
The performance-enhancing index on the expression table is very memory intensive due to the use of string sample identifiers, to match their occurrence in the sample metadata tables.

This is easily mitigated by the addition of a table of integer identifiers for samples, and usage of an extra int column rather than string column. There will be slight change to the usage pattern, whereby the optimized SELECT queries must first lookup the integer identifier (this is preferred to using a JOIN on the new table, for possible performance reasons).

jimmymathews:
This should also include inserting the values for the extra column at record creation time (data import).
Then the auxiliary lookup table can be replaced with a range definition table for each specimen, and the queries replaced with range-delimited versions.

PR #246, opened 2023-11-14
Slim down big table indexing

jimmymathews:
Addresses #226 .

- Adds a new table for tracking scope ranges.
- Converts the former `source_specimen` column on `expression_quantification` to a `SERIAL` integer.
- Makes tabular import keep track of ranges per-specimen in the new `range_definitions` table.
- Updates the "optimized" sparse matrix query to use the ranges rather than the former huge index.

Also:
- Deprecates the modify-constraints CLI entrypoint (only used internally now).
- Deprecates the expression indexing module, CLI entrypoint, etc.


jimmymathews:
Will just wait to merge until successful dataset import using the new system.

PR #264, opened 2023-12-01
il2 importance analysis

### LINKED ISSUE #263: Add importance score evaluations to analysis_replication
Using the web portal, we were able to find some interesting correlations between certain phenotypes and their over- or under-representation in the top 100 most important structures per specimen. This was done by manual inspection, so it's time we formalized them inside `analysis_replication/`.

CarlinLiao:
Merge to close #263.

### DIFF
--- a/analysis_replication/accessors.py
+++ b/analysis_replication/accessors.py
@@ -69,7 +69,7 @@ class DataAccessor:
         if isinstance(phenotype_names, str):
             phenotype_names = [phenotype_names]
         conjunction_criteria = self._conjunction_phenotype_criteria(phenotype_names)
-        all_name = ' and '.join([self._name_phenotype(p) for p in phenotype_names])
+        all_name = self.name_for_all_phenotypes(phenotype_names)
         conjunction_counts_series = self._get_counts_series(conjunction_criteria, all_name)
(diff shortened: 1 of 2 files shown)


CarlinLiao:
I've implemented my addition in the pattern of the other analyses, but I'm getting a small error.

One of the custom phenotypes I'm using is a mixture of "simple" and "complex" phenotypes, ["Tumor", "MHCI"]. This causes an error because the `phenotype-criteria` API endpoint appears to be only set up to handle "complex" phenotypes, as "Tumor" is processed fine but "MHCI" causes a 500 internal server error. I can work around this, but I would think that this isn't necessarily desired behavior.

CarlinLiao:
Once the above issue is resolved, this is ready to merge.

jimmymathews:
Look at the other examples, for non-named phenotypes you have to specify a dict of `postive_markers` list and `negative_markers` list.

jimmymathews:
Also the sequence of results computed here is supposed to match the manuscript narrative, not just list bulk results. So we should select the ones notable enough to have a sentence about them (and also write that sentence), including in this reproduction script in the same order.

jimmymathews:
Looks good.

PR #267, opened 2023-12-04
adjust specimen allocation

### LINKED ISSUE #266: Stabilize GNN training/validation/test set creation when given bimodal specimen sizes
Some studies (e.g., urothelial) contain both whole slide images (WSIs) and tissue microarrays(?). They're treated identically when being allocated to the training, validation, and test sets, so by chance the WSIs could all be allocated to the test set, even though they contain many more ROIs than the microarray specimens. We should stabilize the way they're allocated so that WSIs take priority during specimen allocation.

CarlinLiao:
To close #266. Adds a condition that if a specimen has 10x the number of ROIs created from it than the median specimen (e.g., in a mixed whole slide/microarray dataset).

All tests pass and in my run classification accuracy increased.

### DIFF
--- a/spatialprofilingtoolbox/cggnn/generate_graphs.py
+++ b/spatialprofilingtoolbox/cggnn/generate_graphs.py
@@ -41,6 +41,10 @@ from spatialprofilingtoolbox.cggnn.util.constants import (
 # 10% of the size of the desired ROI size.
 MIN_THRESHOLD_TO_CREATE_ROI = 0.1
 
+# Set the factor for a specimen to be considered "big" relative to the median number of ROIs
+# generated per specimen, e.g., if a study has both whole slide images and tissue microarrays.
+BIG_SPECIMEN_FACTOR = 10
+
 
 def generate_graphs(
     df_cell: DataFrame,
@@ -449,7 +453,11 @@ def finalize_graph_metadata(
     features_to_use: list[str] | None,
     output_directory: str | None = None,
 ) -> list[GraphData]:
-    """Split into train/validation/test sets and associate other metadata with the graphs."""
+    """Split into train/validation/test sets and associate other metadata with the graphs.
+
+    If there's a mix of whole slide images (WSIs) and tissue microarrays (TMAs), this method
+    allocates WSIs first, identified by how many ROIs are created from each image.
+    """
     graphs_by_label_and_specimen = _split_graphs_by_label_and_specimen(graphs_by_specimen, df_label)
     specimen_to_set = _split_rois(graphs_by_label_and_specimen, p_validation, p_test)
     graphs_data = _assemble_graph_data(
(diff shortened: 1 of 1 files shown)


jimmymathews:
Under the condition you mention, what follows? Such specimens are always part of training?

jimmymathews:
I request that you remove the comment lines and replace whatever they are supposed to convey with code functionality and semantic symbols/identifiers. E.g. `# Shuffle the specimen order, but give large specimens priority` may become 

```py
def shuffle_specimen_order(..., shuffle_priority='large'):
   ...
```

Also these lines:

```py
# Set the factor for a specimen to be considered "big" relative to the median number of ROIs
# generated per specimen, e.g., if a study has both whole slide images and tissue microarrays.
```

are problematic because they implicitly promise behavior that is not guaranteed. What is guaranteed is only in the formal code structure.

CarlinLiao:
Finished the sentence fragment from earlier.

For the second part, do you simply mean that "big" should be explicitly defined? Or to exclude the example? The behavior is guaranteed so long as there are very big and/or very small specimens in the dataset.

jimmymathews:
I mean that if the implementation changes, there is no guarantee that the comment will too.

CarlinLiao:
I see. Is there a way to programmatically add documentation to constants? I think these start-of-file constants benefit from explanations about their utility, as the reason why they're defined prior is so that a maintainer can change them to suit their needs or newer discoveries, like in this situation where I set some arbitrary threshold for a "big" image that may change with more experience, but shouldn't be exposed as a parameter.

jimmymathews:
Parameter exposure is a bit of a different issue.
This is just about my hard rule against using inline comments `# ...` in virtually all circumstances.
It is too much freedom, one writes for oneself in such comments without leaving clear instructions to future developers about how to keep the comment (uncontrolled english text) in sync with the actual code behavior.

jimmymathews:
As soon as you start making it clear how the comment contents correspond to variables and functions in source code, you may as well just write the meaning directly into the source code.

You use `BIG_SPECIMEN_FACTOR` once. Can you make it clear that one is supposed to

```py
# Set the factor for a specimen to be considered "big" relative to the median number of ROIs
# generated per specimen, e.g., if a study has both whole slide images and tissue microarrays.
```

by refactoring that 1 usage?

jimmymathews:
This issue is raised [here](https://softwareengineering.stackexchange.com/questions/254978/whats-wrong-with-comments-that-explain-complex-code). The one [answer](https://softwareengineering.stackexchange.com/a/255019) that is highly ranked and defends code comments does so in the tricky case where the correct code is nevertheless misleading. This is what I have done in what might be the only place in the SPT codebase where I use an inline comment:

https://github.com/nadeemlab/SPT/blob/0b2ee983897c0d0ee1402d41013722aef4013eb7/spatialprofilingtoolbox/apiserver/app/main.py#L71-L72

Here the parseable code as written might be perceived as incorrect for a specific unusual reason.

CarlinLiao:
Changes made as requested. All tests pass.

Issue #272, opened 2023-12-08
Abstract out graph deep learning functionality to an external plugin system


CarlinLiao:
This would be implemented as follows
1. The `cggnn` submodule would be renamed to `graphs` to reflect the change in its responsibilities and disambiguate from the `cg-gnn` model. (This would also future-proof against a potential expansion to creating tissue/region graphs in addition to cell graphs that I think we'd want to do, but that's far outside the scope of this issue.)
2. This repository, SPT, would reduce graph functionality in the `graphs` submodule to the following:
   * **Output graphs**: Given a configuration with controlled fields like `graph_specs.yml` (i.e., an API call), output a data artifact with consistent structure to file that contains cell graphs constructed by SPT.
   * **Upload transformations**: Upload consistent output (presumed to be transformed versions of the output graphs, e.g., by a deep learning model), which includes importance scores per histological structure, to the database.
   * Other functions that can utilize the consistent input and output to do simple stuff with, like graph plotting.
3. Graph deep learning functionality would be abstracted and moved to an external plugin system.
   * Each plugin would have
      * A filled-out but study-agnostic version of `graph_specs.yml`, so it can be applied to any study uploaded by SPT.
      * A Docker container that would set up their environment to run their model, run their model on the data artifact, and output results in the format that SPT expects so they can be uploaded to SPT (importance scores to start with, but potentially more stuff like model checkpoints later down the line).
   * These Docker containers would be derived from a template GitHub repo that will be created to close this issue.
4. SPT `workflow` would be modified to get pinned versions of each plugin's Docker container to run the deep learning workflow(s).
   * (A later update will add functionality to `db` that triggers deep learning runs on upload of new studies in a similar fashion.)

Addition of new graph deep learning plugins to SPT would be done in three steps:
1. **Exploratory stage**: a user would evaluate if their deep learning model is good or not by manually using the output and upload functions described in the prior section and examining their results locally or in an SPT instance after upload.
2. **Proposal stage**: If a user determines that their graph model is a good candidate for inclusion into SPT as a default graph deep learning plugin that should be run on every new study imported into SPT, they would containerize their model according to the Docker template defined for the external plugin system, upload it to a GitHub repo or equivalent, and open an issue or pull request for SPT to include their plugin.
3. **Inclusion stage**: The SPT maintainers would quality check the proposed plugin and, if found to be worth including, upload the container to the SPT Docker page and modify SPT code to pull down and use that Docker image by default in SPT.

Implementation plan.
[...]

PR #273, opened 2023-12-11
Compartmentalize DGL and torch usage to `histocartography` and `cg_gnn_helper`

### LINKED ISSUE #271: Compartmentalize dependency on DGL and torch to cg-gnn model
We currently construct cell graphs using DGL, but it's used relatively rarely compared to something like saving a feature and adjacency matrix in pytorch or numpy, even in the GNN space. It's simple enough to convert a DGLGraph to matrices and vice versa, so I propose we refactor the cggnn module such that use of DGL is localized to when data is passed into the deep learning model only, i.e., replace DGLGraphs during graph creation and plotting with matrix representations.

This would also help with #268, as that's an example of a GNN(-Transformer) model that doesn't use DGL at all.

CarlinLiao:
To close #271. Future changes will extricate `histocartography` and `cg_gnn_helper` into a separate plugin as described in #272.


CarlinLiao:
All tests pass.

jimmymathews:
In testing I encountered an issue I have encountered before, where the prereqs image has outdated layers (apt index/servers list I think). So deal with this I added a "NOCACHE=1" option to the Makefile, which one can use in such situations.

CarlinLiao:
Could you release this to pip?

jimmymathews:
[Done](https://pypi.org/project/spatialprofilingtoolbox/#history).

PR #276, opened 2023-12-14
Dockerize `spt cggnn train`

### LINKED ISSUE #275: Dockerization of the `cg-gnn` model
Pursuant to step 2 of #272's implementation plan, we should convert `cggnn` to use a Dockerized version of the `cg-gnn` model. Steps as follows:
1. Refactor the `cggnn` submodule to provide graphs and intake deep learning models using a model-agnostic I/O API (already done)
2. Excise and Dockerize the `cg-gnn` model from the SPT repo (undoing #257) and refactor it to use the new `cggnn` I/O API
3. Modify existing tests for `cggnn` and `workflow` to pass using the Docker container

Dockerizing `cg-gnn` will pave the way for implementing the plugin process described in #272 and have an immediate benefit of allowing us to remove torch and DGL from our SPT development Docker image, speeding up test suite runs dramatically.

CarlinLiao:
Closes #275. This PR will remain a draft until #273 is accepted and merged, a new SPT version is released using it, and the functionality of this branch and the Docker image it uses is testable.


CarlinLiao:
I am hard stuck on trying to run Docker (the spt-cg-gnn plugin) inside Nextflow inside Docker (the test suite harness).

When I have Nextflow run `docker run`, it finds the correct docker image (nadeemlab/spt-cg-gnn:latest) but the volume I passed from Nextflow is nowhere to be found, even if I copy the files in the Nextflow process script before running docker.

When I have use the Nextflow `container` field inside Docker, the files Nextflow prepares from inputs are easily found, but it's in the wrong Docker container. `\$(docker inspect --format='{{.Config.Image}}' \$(cat /etc/hostname))` returns `nadeemlab-development/spt-development:latest`, not `nadeemlab/spt-cg-gnn:latest`.

Is it actually possible to run docker-in-nextflow-in-docker properly? Do we have to do this? The standard use case would stop at Nextflow running Docker, without the need to wrap another Docker layer around it all.

jimmymathews:
I assume you don't mean literally using `docker run` inside a Nextflow process. That is not likely to work. It is not possible as far as I know to run docker inside a docker-contained Nextflow process, or any other docker-contained process.

Nextflow is itself an orchestration software. It is capable of doing allocation in a variety of environments, and it is responsible for summoning docker containers when asked. Docker compose is a lower-tier orchestration software, meant for prototypes in a simpler environment.

Our plan for a plugin system was to inform Nextflow of the correct (Singularity) container image to use for a specific process. The tests of such a system don't fit the pattern of our current integration tests, since these existing tests are based on a docker composition. We will have to write new tests that are shell scripts invoking Nextflow. Some Nextflow docs about this can be found [here](https://www.nextflow.io/docs/latest/container.html#id24).

CarlinLiao:
It is very much possible to run [Docker-in-Docker](https://www.docker.com/blog/docker-can-now-run-within-docker/). Docker in Nextflow in Docker may be a different story.

I've tried
* [Using the Docker run method suggested by nextflow](https://www.nextflow.io/docs/latest/container.html#docker), as you pointed to
* Having the Nextflow process shell script call Docker run
* Having the Nextflow process shell script call a python process that calls docker run

None of the above work, but they don't break in anexpected way. Docker run does run, it just errors when trying to find the file I'm trying to provide it because I can't find a way to pass data from the Nextflow context into the Docker context.

I'll push my sketch implementations for review tomorrow.

CarlinLiao:
To avoid a Docker-in-Docker situation, we decided that we would adjust the cggnn workflow test to run outside of the Dockerized test harness. It'll still query the (Dockerized) small test database, but its test script will now run outside of Docker. This will require changing the Makefile that runs the test suite.

Once this test is implemented and all tests pass, this PR will be ready for merge.

spatialprofilingtoolbox/workflow/assets/cggnn.nf
```diff
@@ -133,6 +135,10 @@ process finalize_graphs {
 process train {
     publishDir '.', mode: 'copy'
 
+    // beforeScript 'export DOCKER_IMAGE=$(if [[ "${cuda}" == "true" ]]; then echo "nadeemlab/spt-cg-gnn:cuda-latest"; else echo "nadeemlab/spt-cg-gnn:latest"; fi)'
+
+    // container "\${DOCKER_IMAGE}"
```
CarlinLiao, line 140: We had some discussion about how we want to implement this train step here, and settled on replacing the container with a very unique string that's replaced with the cuda or non-cuda version of container depending on your inputs to `spt workflow configure`. Does that need to happen in this PR, or can that wait for later? Either when closing #272 or as a later feature enhancement, instead letting the container default to non-CUDA for now?

build/workflow/Makefile
```diff
@@ -81,6 +83,27 @@ ${TESTS}: setup-testing
 >@${MESSAGE} end "Passed." "Failed."
 >@rm -f status_code
 
+${STANDALONE_MODULE_TEST}: ${MODULE_TESTS}
+>@test_target=$$(echo $@ | sed 's/..\/..\/${TEST_LOCATION}\/${MODULE_NAME}\///g') ; \
+    test_script=$${test_target/test-/} ; \
```
CarlinLiao, line 88: I copied this from the `${TESTS}` rule but I don't understand it. I don't think we have any test files that start with `test-` instead of `test_`?

jimmymathews: The line
```make
MODULE_TESTS := $(foreach test_file,$(MODULE_TEST_FILES),test-$(test_file))
```
prepends `test-` in order to create a **target** name. `make` targets are not the same as files, and in this case it is especially important that the test file not be confused with the test target. The reason is that `make` will specially interpret target names which coincide with file names, as the names of files that the recipe portion (the script) is supposed to create.

The lines you are referring to just recover the test file name from the target name. It's roundabout and ugly because I'm not an expert in make string processing.

jimmymathews: If you'd like, in your case you could hard-code a single target and recipe, removing the need for the string name processing.

jimmymathews: Like:
```make
standalone-module-test-1: setup-testing
>@bash standalone_module_tests/my_special_test.sh; \
    echo "$$?" > status_code;
...
```

then add to the definition of the test list:

```make
TESTS := ${UNIT_TESTS} ${MODULE_TESTS} standalone-module-test-1
```

spatialprofilingtoolbox/workflow/assets/cggnn.nf
jimmymathews: I think it's part of the current issue. It means establishing the pattern whereby a docker repository name is configurable at runtime and used in this specific place. Check the Nextflow docs to see if it's possible to do process-specific docker container requests in the nextflow.config, so that we wouldn't have to alter the `.nf` definition. Otherwise use something like

```
// container ZZZ_GRAPH_PROCESSING_CONTAINER_NAME_ZZZ
```
in the `.nf` file and
```py
re.sub(special_string, container_name, nf_file_contents)`
```
during generation of the `.nf` file.

CarlinLiao:
[FYI from the Nextflow docs](https://www.nextflow.io/docs/latest/process.html#beforescript)

> When combined with the [container directive](https://www.nextflow.io/docs/latest/process.html#process-container), the beforeScript will be executed outside the specified container. In other words, the beforeScript is always executed in the host environment.

jimmymathews:
Nice, that it is the first part that is needed from that line. Next what is needed is to ensure that the result of that script is actually exported to the level of the process definition (including container directive). Since this note indicates support in precisely the condition we face ourselves with, I'm willing to bet that it does work as expected. Most likely a configurable container is the reason for this behavior.

CarlinLiao:
[Something that I'm not a fan of is this.](https://stackoverflow.com/questions/75321637/using-docker-for-only-some-processes-in-nextflow#comment132969074_75322219)

> I can confirm that (a) it is not possible to have only some of the processes running within Docker containers and that (b) this has been discussed among the developers and there are no plans to have this "feature". Also worth mentioning is that in some executors (cloud computing such as using Google Batch), a container image is required (and will be used) regardless of docker.enabled = true or -with-docker.

So for `cggnn.nf` I have to specify a Docker container to run every single process in, I can't only have the train process run in a container and the rest on native. How would you like to proceed?

CarlinLiao:
Figured that issue out. Since each process needs a Docker container, I can just pass the development image to all non-train files, and provide the network that's already built for workflow testing to all of them.

This fixes everything, all that's left is to debug the cg-gnn image.

CarlinLiao:
All tests pass. But the new non-Dockerized test's output is a little wonky.

```
workflow ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Setup env.     (19s)     
┃ proximity pipeline ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Passed.        (35s)     
┃ umap plot creation ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Passed.        (37s)     
┃ cggnn ┈┈┈cat: status_code: No such file or directory
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ /home/carlin/projects/SPT/build/build_scripts/verbose_command_wrapper.sh: line 88: [: -gt: unary operator expected
Passed.        (46s)     
┗━ ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ Down.          (1s) 
```

build/workflow/Makefile
```diff
@@ -81,6 +83,29 @@ ${TESTS}: setup-testing
 >@${MESSAGE} end "Passed." "Failed."
 >@rm -f status_code
 
+${STANDALONE_MODULE_TEST}: ${MODULE_TESTS}
```
jimmymathews, line 86: MODULE_TESTS should not be a dependency of your standalone test.

build/workflow/Makefile
```diff
@@ -81,6 +83,29 @@ ${TESTS}: setup-testing
...
+>@test_target=$$(echo $@ | sed 's/..\/..\/${TEST_LOCATION}\/${MODULE_NAME}\///g') ; \
+    test_script=$${test_target/test-/} ; \
+    test_name=$${test_script/module_tests\/no_docker_test_/} ; \
+    test_name=$${test_name/unit_tests\/no_docker_test_/} ; \
+    extension=$$(echo $$test_name | grep -o "\.sh\|\.py$$" ) ; \
+    test_name=$${test_name/$$extension/} ; \
+    test_name=$$(echo "$$test_name" | tr '_' ' ' ) ; \
```
jimmymathews, line 93: These lines were meant to recover the test name from the test target name, but you have named STANDALONE_MODULE_TEST with the literal name of the test file. This creates the conflict between target names and file names that I mentioned elsewhere, and it also means that the above lines are mostly unnecessary.

build/workflow/Makefile
```diff
@@ -81,6 +83,29 @@ ${TESTS}: setup-testing
...
+>@test_target=$$(echo $@ | sed 's/..\/..\/${TEST_LOCATION}\/${MODULE_NAME}\///g') ; \
+    test_script=$${test_target/test-/} ; \
+    test_name=$${test_script/module_tests\/no_docker_test_/} ; \
+    filename="module_tests/$$(basename $$test_script)" \
+    extension=$${filename##*.} ; \
+    executor=none-cmd ; \
+    if [[ "$$extension" == "py" ]]; then executor=python; fi ; \
+    if [[ "$$extension" == "sh" ]]; then executor=bash; fi ; \
+    cd ../../${TEST_LOCATION}/${MODULE_NAME}/ ; \
+    $$executor $$filename ${DOCKER_ORG_NAME}-development/${DOCKER_REPO_PREFIX}-development:latest ; \
```
jimmymathews, line 104: You don't need to programmatically discover the appropriate executor for this one test -- it's a .sh shell file. You can just use "bash".

build/workflow/Makefile
```diff
@@ -81,6 +83,29 @@ ${TESTS}: setup-testing
...
+    echo "$$?" > status_code
```
jimmymathews, line 105: This line probably isn't running, hence no status_code, hence the error you observed.

CarlinLiao: I can confirm that it runs since it produces Nextflow file output.

jimmymathews: No, the error you reported is proof that this specific line did not run.

```sh
cat: status_code: No such file or directory
```

jimmymathews: There is also a predecessor error:

```
Output importance file (     149) is not the expected length (149).
```

So, formatting of the word count size string.

jimmymathews: I think `status_code` was not being created because you previously `cd` into an unexpected directory before running the test. So it is created in some other directory.

test/workflow/module_tests/no_docker_test_cggnn.sh
```diff
@@ -1,5 +1,11 @@
+# Set up a virtual environment to run this test
+python -m venv env
+source env/bin/activate
+pip install ../..[workflow]
```
jimmymathews, line 4: We really shouldn't be creating inception virtual environments... installing from source then abandoning it.
The spt command should clearly be run in the development image, just like in all the other tests.
The nextflow run is what needs to be run outside, and now we have a new development environment dependency: nextflow. Previously nextflow was entirely containerized, but for this specific test we aren't doing that.

jimmymathews: (The install of spatialprofilingtoolbox and extras doesn't properly belong to the test functionality and wastes time.)

CarlinLiao:
Thank you for the notes on the build and test process. I finished your changes and now all tests pass without error in the tests or the scaffolding that runs them. I believe this is now fully ready to merge.

test/workflow/module_tests/no_docker_test_cggnn.sh
```diff
@@ -0,0 +1,81 @@
...
+    else
+        echo "$overlap_count% of the most important histological structures match the reference."
+    fi
+}
+
+function clean() {
+    rm -f .nextflow.log*; rm -rf .nextflow/; rm -f configure.sh; rm -f run.sh; rm -f main.nf; rm -f nextflow.config; rm -f top_100_structures.txt; rm -f top_100_reference.txt; rm -rf results/
+}
+
+nextflow run . -with-docker ${docker_image}
```
CarlinLiao, line 74: Recall that `spt workflow configure` outputs a shell file that says `nextflow run .`. We should either remove it or change it so that it prints out `nextflow run . -with-docker ${docker_image}` if the workflow is cg-gnn.

If we want to keep it, we would need to feed the name of the docker image that's running everything _but_ cg-gnn to `spt workflow configure`. #272 and #278 want the inputs to graphs to be centralized into configuration files, so for the cg-gnn workflow test to work, we'd need to pipe the name of the docker image into the configuration file that runs `spt workflow configure`... This seems too convoluted to me but the alternative is to just assume the name of the Docker image in a static configuration file we write, or carve out another special CLI argument for cg-gnn which goes against what #278 is looking to do.

CarlinLiao: Nevermind, I can bake these options into `nextflow.config.jinja`.

CarlinLiao: Although part of the issue is still there in that unless we create some kind of CLI that feeds the default image and network into a config template to create the actual config file, we're hardcoding the Docker image and network names into a config file whereas before they were dynamically generated by the Makefile. (This concern will make more sense in the #272 branch.)

CarlinLiao:
I accidentally merged this pull request just now, and there doesn't appear to be an easy way to unmerge it without reverting history.

jimmymathews (commented):
Checked that pytorch is no longer required here, and tests all pass except the new standalone one for reproducibility reasons it seems. The system of using the cg-gnn image released on dockerhub seems to work.

jimmymathews:
Manually squash merged into main.

test/workflow/module_tests/no_docker_test_cggnn.sh
jimmymathews: Yes this configurability should be sorted out eventually. For now let's not let it block merge however.

PR #280, opened 2023-12-17
Save adjacency matrices separately

### LINKED ISSUE #279: scipy sparse matrix compatibility issues
In the current version of the cg-gnn graph saving and loading, graphs are created as adjacency matrices, which are pickled as scipy arrays. This causes compatibility issues because different versions of scipy make tweaks to the sparse matrix datatype that can cause loading to error if the scipy version used to save graphs is even slightly different than the version used to load the graphs (e.g., in SPT versus the spt-cg-gnn Docker image). We should implement graph saving and loading such that the adjacency arrays are saved separately from the rest of the graph data, which uses much simpler datatypes that don't have this issue.

CarlinLiao:
To close #279. All tests pass. This needs to be closed and a new version of spatialprofilingtoolbox released to pip to fix the [nadeemlab/spt-cg-gnn](https://github.com/nadeemlab/spt-cg-gnn) [docker image.](https://hub.docker.com/r/nadeemlab/spt-cg-gnn)


jimmymathews:
v0.17.4 released.

Issue #278, opened 2023-12-14
Refactor `spt workflow configure` to accept a generic configuration file instead of CLI arguments


CarlinLiao:
As suggested by #247 and necessitated by #272, allowing the user to swap graph processing plugins that likely have their own set of input configurations. (There was also discussion of removing configurability of the models altogether, but that would also be in line with this issue as the current options exposed for the `cggnn` workflow would need to be removed in that case.

The configuration YAML could look something like this, with a generic [settings] section that applies to all workflows and sections for specific workflows or graph models.

```
[settings]
db_config_file = ...
executor = ...
workflow = ...

[db_visitor]
file_manifest_path = ...

[graphs/cggnn]
strata = ...

[graphs/gnn-transformer]
depth = ...
```

The settings could but wouldn't need to be saved to the same config file, as the logic would only use the relevant section for the task at hand, and they're all named distinctly.

jimmymathews:
This feature is valuable, but let's not confound it with the declarative configuration that was planned for the graph processing plugins. That (probably substantially slimmed down) configuration is perhaps better considered as supporting a single function call. Plugin developers should probably not need to be aware of `spt workflow ...`, with its other more low-level options.

PR #281, opened 2023-12-29
Convert cggnn/graphs and workflow configure to use config files

### LINKED ISSUE #272: Abstract out graph deep learning functionality to an external plugin system

CarlinLiao:
This closes #272 and #278 by
* Renaming the `cggnn` submodule to `graphs`
* Converting most `graphs` functionality to control by config file and not CLI args
* Converting `workflow configure` to be controlled by config file and not CLI args

#276 and [spt-plugin#3](https://github.com/nadeemlab/spt-plugin/pull/3) are necessary prerequisites for this PR as #276 is still dependent on an older version of spt-cg-gnn Docker image that will need to be updated once spt-plugin#3 is accepted. Until those are merged and that Docker image updated, this will remain an untested draft PR.


CarlinLiao:
This was also accidentally merged by me at the same time as #276, but this wasn't a PR that was ready to merge. I think what happened is that I pressed the wrong button in the VS Code version control UI (but I didn't think that it was possible to merge pull requests in VS Code). Checking resolution methods...

CarlinLiao:
I suspect but can't confirm what happened was that
1. my local SPT repo was sitting in the issue272 branch
2. I used the VS Code UI to switch to main branch and sync for new incoming commits and upload outgoing commits, but the switch only switched my branch reference and not repo state,
3. so all the commits from issue272 got uploaded to main as if rebasing.

Consequently, there's no single merge commit to undo, all of the commits from issue275 and issue272 are in main as if I commit them to the main branch directly as I was working on them. Conveniently, however, the branches happened to be worked on serially over the last few weeks so it should be easy to split: every commit [after this one](https://github.com/nadeemlab/SPT/commit/04e2feffb4bbf315b83843b9f489bca9c61bb182) is from issue275 until [this one](https://github.com/nadeemlab/SPT/commit/98bec1757ca9fccd1ad26d1018b71f9033fae576), after which it's commits for issue272.

jimmymathews:
Yes, that it what I was going to recommend (not sure about whether `--hard` is necessary though).
I believe the issue branches can be worked on normally after this. I'll give it try.

jimmymathews:
It looks OK to me.

CarlinLiao:
Thank you for fixing my mistake, it looks good on my end as well. I don't see a method to re-open these pull requests, though I may not have the necessary permissions for that.

jimmymathews:
Hm well the branches are still there (e.g. `issue272`) so I think you can just make a new PR.

PR #287, opened 2024-01-13
Make `spt workflow configure` support both Docker and Singularity

### LINKED ISSUE #286: Expand workflow configure support for singularity
Some HPCs (like ours) ban the use of Docker, but permit the use of Singularity.

Presently, `spt workflow configure workflow='cg-gnn'` requires the use of Docker. We'd like it to support using Singularity instead of Docker. [Conveniently, Nextflow has been integrating more Singularity support and Singularity has been integrating more Docker support, so this shouldn't be too hard a lift.](https://www.nextflow.io/docs/latest/singularity.html)

While we're at it, we should also revisit other places where Singularity is used in `spt workflow configure` and integrate them into this new Singularity run option, if convenient.

CarlinLiao:
Closes #286, technically. `sif-file` and `docker` image commands are merged together in the `workflow configure` configuration file. All existing tests pass, and I think this would work on a machine that has both Singularity and Docker installed, but I've yet to test it on a machine with Singularity installed.


spatialprofilingtoolbox/workflow/assets/.workflow.config.template
```diff
@@ -1,8 +1,9 @@
 [general]
 db_config_file = path/to/db.config
 executor = local
-sif_file = path/to/sif/file
 excluded_host = None
+orchestrator = local
```
jimmymathews, line 5: I think "orchestrator" may be a bit of a misnomer here.
Nextflow is the orchestrator. The options for this field are "local", "docker", or "singularity", and these all describe a type of platform that the orchestrator calls upon for running individual processes.
Perhaps "process_run_delegation"?

spatialprofilingtoolbox/workflow/assets/.workflow.config.template
```diff
@@ -1,8 +1,9 @@
...
+image = name/of/or/path/to/docker/or/singularity/image/if/not/local
```
jimmymathews, line 6: The template text here suggests that a more descriptive name than "image" might be desired.
How about:
`image_name_or_path`

spatialprofilingtoolbox/workflow/assets/.workflow.config.template
CarlinLiao: This is fair. I struggled to disambiguate between the "executor" and the "orchestrator," and per your comments, Docker/Singularity/local is yet another distinct concept. Maybe something more apt would be [a "container" platform, as SingularityCE describes itself in its docs.](https://docs.sylabs.io/guides/latest/user-guide/introduction.html)

CarlinLiao:
I've been investigating different ways to integrate Singularity into our processes and possibly our test suite.

Singularity can create SIF files from local Docker images, but as I understand it Nextflow using Singularity cannot. So, if we want the `cg-gnn` workflow to work with Singularity, we need to make available not just the `spt-cg-gnn` Docker container, but an equivalent of the `spt-development` or at least the `spt-workflow` Docker container on Docker Hub that Nextflow-Singularity can pull from and convert.
* Alternatively, we add some lines to the `run.sh` file such that it converts a local image to SIF file before `nextflow run .`.

Should this part run during `spt workflow configure`, `nextflow run .`/`sh run.sh`, or should they assume that a SIF file already exists? If the latter, should we create an SPT CLI command that creates SIF images or just assume the reader knows how to do that? Unfortunately, it can't go inside the Nextflow script as the images are a necessary prerequisite. (Personally, I prefer integrating SIF creation into the workflow as we've already pinned two specific versions of `spt-cg-gnn` to be used, which having the user provide SIFs themselves bypasses, but I can understand why we could also want them to already be created since Docker-to-SIF is a large operation that could end up being repeated unnecessarily.)

Furthermore, I don't think we can do this on the HPC because as I understand it, Singularity can't convert Docker images to SIF without Docker being installed on that same machine. We can work around this in two ways:
1. Make a canonical part of the build process using a machine (like our laptops) that have both Docker and Singularity installed (and maybe even a CUDA GPU may also be required in the case of a CUDA-utilizing Docker image like `spt-cg-gnn:cuda` but this I'm less sure about) to build the SIF file, then upload it onto something like Box so it can be downloaded onto the HPC.
   * This would require that we add a method to specify paths to SIFs instead of just the Docker tag to use.
3. Use [`singularity build --remote`](https://docs.sylabs.io/guides/3.1/user-guide/singularity_and_docker.html#building-containers-remotely), which uses a remote service to build images from Docker Hub into SIF files that are then downloaded.
   * As far as I understand it, this is free but requires a Sylabs account so you can generate the authentication token required to build on their remote.

Building these options into our test suite would add more complications.
* Singularity would become one of the development and test requirements.
* To integrate this into our test suite, we would have to add a new test that's a modification of the `cg-gnn` workflow test to use Singularity instead of Docker.
* Since we don't want to rely on SPT images from Docker Hub as that would mean using SPT code that's outdated compared to what the test suite is testing, we can't use option 2 in testing unless we upload a new `spt-workflow` image to Docker Hub with each test.

CarlinLiao:
Following in-person discussion,
* we concluded that we will not target inclusion of Singularity in the test suite, obviating some of the points I raised earlier
* Singularity can convert images from Docker Hub to SIF without Docker being installed on the system
* we'll re-introduce the spt-workflow image in this PR
* we should avoid depending on an additional system, i.e., Sylabs
   * but we don't need to have the user create SIFs manually since Singularity will just pull images from Docker Hub and convert them as needed

CarlinLiao:
We've also decided to deprecate LSF support, in part to clean up the workflow configure options.

spatialprofilingtoolbox/workflow/assets/.workflow.config.template
jimmymathews: As discussed, how about: `container_platform` as `docker`, `singularity`, or `none`, with default value `none`.

CarlinLiao:
Upon discussion, we want to make it so that we pin not only the cg-gnn training image, but also the image used to run the rest of the workflow if the user chooses run use Singularity or Docker (for now, this is just `nadeemlab/spt` but in the future different workflows could have their own images, like `nadeemlab/spt-phenotype-proximity`. The `image` configuration option will be changed to a `tag` to allow the user to override which _version_ of the image is used, but by default workflow will use `latest`.

CarlinLiao:
I still haven't gotten it to work. I did a little more testing on my local machine which has both Singularity and Docker installed and found that, for whatever reason, when using Docker as the containerization platform the beforeScript section correctly assembles the `DOCKER_IMAGE` environment variable, but not when using Singularity.

I think the simplest solution will be to do the cuda calculation before Nextflow starts running by having `spt workflow configure` assemble the training image name instead. That's presuming that the issue is with beforeScript and not the `container` flag in process not having access to any outside variables at all...

CarlinLiao:
Now that I think about it, I found no evidence of beforeScript running when using Singularity before the error, so perhaps while Docker behaves correctly and runs beforeScript before container, maybe when using Singularity beforeScript is run after container (but before the, well, script) or not at all.

jimmymathews:
Yeah since it is a very small usage of the beforeScript functionality, perhaps it is better to move this functionality into the configuration prep alongside everything else like it.

jimmymathews:
Accidentally closed sorry.

CarlinLiao:
This seems to have fixed it but I'm doing an HPC run to be sure. We should probably report this bug to Nextflow but I can't be bothered to pull together the documentation.

CarlinLiao:
The improperly resolving Docker image problem was fixed, but I'm encountering a new issue: when Singularity is trying to convert the training Docker image to SIF, it errors and complains that there's no space left on device. This is despite me running the operation in /nadeemlab/ which has 15 *terabytes* of free space.

Given the error message, it might be referring to the HPC itself since that's where my conda environment lives. That has 12G of free space right now, which could be the culprit. 12G is still a ton though... just how big are SIFs?

CarlinLiao:
Upon further investigation, `NXF_SINGULARITY_CACHEDIR` should be set to where we have a lot of space, as that's where Nextflow-Singularity will cache images after converting them to SIF. This clobbers usage of the `SINGULARITY_CACHEDIR` environmental variable used when running Singularity outside of Nextflow.

But another part of the process that uses a lot of memory is the temp directory used by Singularity as it's converting an image. By default it's `/tmp/`, but on our HPC that folder doesn't get allocated more than 12G, which is what's needed for the conversion. To change it from default, we need to set `SINGULARITY_TMPDIR`, which doesn't have a Nextflow equivalent.

See [here](https://docs.sylabs.io/guides/3.5/user-guide/appendix.html) for more on Singularity environmental variables, and [here](https://github.com/nextflow-io/nextflow/issues/2685) for how they interact with Nextflow.

Should we add options to change these environmental variables in `spt workflow config`, or let users figure it out on their own with guidance from our documentation? If we were to add them, I think they'd have to be baked into `run.sh`, preventing people from using `nextflow run .` directly.

jimmymathews:
Thanks for investigating this.
I think we are not planning to become experts in nextflow/singularity environments, in the sense that a user had better prefer to consult the nextflow/singularity docs for these specific things rather than use whatever we bake into the configuration. So let us just document what we can and refer to the upstream library docs.

CarlinLiao:
I ended up releasing a new cg-gnn pip version that relaxed the python requirement slightly and updated the README. I've also set the singularity tmp directory to the large file store volume. Things are almost there, but I keep getting this error with the permissions in the large file store; I presume this is an issue with our system and not Nextflow or Singularity, but I'm not sure what to do about it. I was able to dodge this issue earlier, but I haven't been able to overcome it this time.

```
$ nextflow run . -resume
N E X T F L O W  ~  version 23.10.1
Launching `./main.nf` [elated_feynman] DSL2 - revision: 3c00f5e95d
[6d/73226d] process > echo_environment_variables [100%] 1 of 1, cached: 1 ✔
[dc/b9900e] process > prep_graph_creation        [100%] 1 of 1, cached: 1 ✔
[40/4448d2] process > create_specimen_graphs (9) [100%] 9 of 9, cached: 9 ✔
[c7/9baba5] process > finalize_graphs            [100%] 1 of 1, cached: 1 ✔
[-        ] process > train                      -
[6d/73226d] process > echo_environment_variables [100%] 1 of 1, cached: 1 ✔
[dc/b9900e] process > prep_graph_creation        [100%] 1 of 1, cached: 1 ✔
[40/4448d2] process > create_specimen_graphs (9) [100%] 9 of 9, cached: 9 ✔
[c7/9baba5] process > finalize_graphs            [100%] 1 of 1, cached: 1 ✔
[-        ] process > train                      -
[-        ] process > upload_importance_scores   -
Pulling Singularity image docker://nadeemlab/spt-cg-gnn:cuda-0.0.2 [cache /nadeem_lab/Carlin/singularity_cache/nadeemlab-spt-cg-gnn-cuda-0.0.2.img]
ERROR ~ Error executing process > 'train'

Caused by:
  Failed to pull singularity image
  command: singularity pull  --name nadeemlab-spt-cg-gnn-cuda-0.0.2.img.pulling.1705632798284 docker://nadeemlab/spt-cg-gnn:cuda-0.0.2 > /dev/null
  status : 255
  message:
    INFO:    Converting OCI blobs to SIF format
    INFO:    Starting build...
    FATAL:   While making image from oci registry: error fetching image to cache: while building SIF from layers: packer failed to pack: while unpacking tmpfs: error unpacking rootfs: chown rootfs: lchown /nadeem_lab/Carlin/singtmp/build-temp-1334923607/rootfs: invalid argument
```

jimmymathews:
I'm pretty sure singularity does fancy things with the file system, which may be blocked by the NFS (network file share) that we use to have all those terabytes of storage space. There are [others with tangentially related issues](https://github.com/apptainer/singularity/issues/2282) with singularity containers on NFS.

jimmymathews:
There are also singularity [docs about NFS limitations](https://docs.sylabs.io/guides/4.0/admin-guide/installation.html#filesystem-support-limitations), which you may have encountered.

jimmymathews:
As a temporary measure we can try to do the SIF image pulling on a separate machine.

CarlinLiao:
I've only done it on the HPC twice or so, but I have a 100% success rate with using `singularity pull` directly but maybe a 10% rate when letting Nextflow do the pull for me. I wonder if we can drop the img files directly into SIngularity's cache instead of modifying the configuration files, just as a backup in case we hit these errors.

CarlinLiao:
I worked around this by clearing out space on the HPC's internal storage (not the external file store) and setting `SINGULARITY_TMPDIR` to be a folder in internal storage even though I was running the workflow in the external file store, and it worked.

That led to discovering that the training process was running much slower than normal because Singularity isn't provided GPUs by Nextflow by default, so I amended the library to support that. It runs properly and is ready to merge provided tests show no breaking changes in the non-Singularity, non-GPU workflow.

jimmymathews:
I also checked on the HPC to do a basic run, seems to work.

PR #288, opened 2024-01-17
prevent package caching in Docker images

### LINKED ISSUE #274: Would adding ` --no-cache-dir` to pip install commands slim down Docker image sizes?
By default, pip caches downloads when installing a new package. That's useful typically but since our Dockerfiles are often tossed, the cached packages end up bloating the size of our images without ever getting used since if we want to renew our Docker image it discards the image and redownloads the wheels, if I understand this correctly. Would changing commands to, say,

```
RUN python -m pip install --no-cache-dir toml
```

reduce our memory footprints in a worthwhile way?

CarlinLiao:
Closes #274. All tests pass. Ideally, I'd like to bring back the `workflow` Docker image in this PR as well to serve #287.


jimmymathews:
The images are indeed smaller. I did a release with the version bumps so this is ready to go as soon as tests pass.

PR #292, opened 2024-02-02
Add support for graph-transformer plugin

### LINKED ISSUE #268: Add tmi2022 GNN model
[A recently released paper](https://doi.org/10.1109/TMI.2022.3176598) includes [a new GNN-Transformer architecture](https://github.com/vkola-lab/tmi2022) that may work better than the CG-GNN currently included in our dataset.

CarlinLiao:
Will close #268.

Per the tests, something seems to have broken with the apiserver initialization that could use a second look over before merging.


CarlinLiao:
Removing that stray import fixed the major apiserver issue, but the test verifying the change to the way importance scores are indexed is still failing.

jimmymathews:
Can you report the logs from the tests that failed?

CarlinLiao:
I don't have them on this computer, it'll have to wait until the evening or tomorrow. It was just the apiserver importance module test though. If I recall correctly the issue was that the data object parsed from the JSON returned from the API query was empty.

I force rebuilt my data loaded images so I don't think that was the issue.

CarlinLiao:
Upon inspection, the current implementation of GraphsAccess could potentially return multiple feature specifications. We decided that we should break out getting one feature specification as a helper function to ensure that it only comes back with one feature specification.

CarlinLiao:
I separated the feature specification fetch in GraphsAccess so it'd always only return one or zero specifications, but I can't figure out why the test fails now.

CarlinLiao:
All tests pass. It took like 10 or 20 minutes, but the total time at the end showed 219393s or 2.5 days, which I can assure you is a slight overestimate. This isn't the first time either.

jimmymathews:
Tests pass for me.

PR #294, opened 2024-02-02
Quick patch to counts_by_signature to handle if one of positive or negative is empty

### LINKED ISSUE #293: `DataAccessor.counts_by_signature` fails if either no positive or no negative markers are provided
The `anonymous-phenotype-counts-fast` API endpoint requires both `positive_marker` and `negative_marker` channels to be submitted, even if one is an empty list. This causes `DataAccessor.counts_by_signature()` to error if you submit, e.g., one `positive_marker` and no `negative_marker`.

We may want to consider editing the API endpoint to make marker submission optional, but for now the quickest solution would just be to ensure that `positive_marker` and `negative_marker` are always submitted by `DataAccessor.counts_by_signature()`, even if one is empty.

CarlinLiao:
Closes #293 by patching DataAccessor.

### DIFF
--- a/analysis_replication/accessors.py
+++ b/analysis_replication/accessors.py
@@ -170,7 +170,13 @@ class DataAccessor:
         df = DataFrame(rows).set_index('sample')
         return concat([self.cohorts, self.all_cells, df], axis=1)
 
-    def counts_by_signature(self, positives, negatives):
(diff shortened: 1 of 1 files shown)


jimmymathews:
I checked that a couple of the analysis scripts still work, will merge.

PR #295, opened 2024-02-05
Counts files to db migration

wooferclaw:
not yet tested


jimmymathews:
I have moved the lines, that actually invoke caching, to the end of the tabular import workflow (main.nf). Now the ondemand containers (start.py) should rarely have to actually do the caching on their own. They still check to see if the cache is present, and recreate it if not, but in normal circumstances this should happen before the service containers are created.

jimmymathews:
All tests pass for me.

jimmymathews:
After this is merged we will need to update spt-dashboard to stop using volumes. We should also do a fresh dataset import using the new functionality.

PR #320, opened 2024-05-09
Make GNN plots available as an SPT CLI command and API endpoint

### LINKED ISSUE #319: Make GNN plots available as an SPT CLI command and API endpoint
This involves moving the current `analysis_replication/gnn_figure/graph_plugin_plots.py` out of `analysis_relication/` and into `spatialprofilingtoolbox/graphs/`, adding a CLI command for it to `scripts/`, and an API endpoint that calls it to `spatialprofilingtoolbox/apiserver/app/main.py`.

CarlinLiao:
To close #319.


CarlinLiao:
Although I wanted to change the database query to be direct instead of through the API server, for expediency's sake + don't-fix-what-ain't-broke, I'll stick with the API server counts implementation and commit a thinner version of DataAccessor for now.

CarlinLiao:
`test/apiserver/unit_tests/record_counts1.txt` and `build/build_scripts/expected_table_counts.txt` are identical. Is this redundancy necessary?

spatialprofilingtoolbox/apiserver/app/main.py
```diff
@@ -375,3 +378,60 @@ async def get_plot_high_resolution(
...
+
+@app.get("/importance-fraction-plot/")
+async def importance_fraction_plot(
+    study: ValidStudy,
+    img_format: str = 'svg',
+) -> StreamingResponse:
+    """Return a plot of the fraction of important cells expressing a given phenotype."""
+    APPROVED_FORMATS = {'png', 'svg'}
+    if img_format not in APPROVED_FORMATS:
+        raise ValueError(f'Image format "{img_format}" not supported.')
```
jimmymathews, line 391: I think this validation might be better implemented in the same system as everything else (FastAPI's recommendation). So validation failure is handled more gracefully by FastAPI.

spatialprofilingtoolbox/db/scripts/upload_sync_small.py
```diff
@@ -0,0 +1,85 @@
...
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        prog='spt db upload-sync-small',
+        description='Synchronize small lists of strings for each study with the database.'
+    )
+    parser.add_argument(
+        'name',
+        help='The name of the table of strings to be synchronized.',
+    )
```
jimmymathews, line 24: I propose that we make this into a mutually-exclusive argument group (see [docs](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group)), or else provide the [`choices`](https://docs.python.org/3/library/argparse.html#choices) argument, so that the names are available for inspection at the CLI.

jimmymathews:
This work entails updating the documentation in spt-data:
https://github.mskcc.org/mathewj2/spt-data/tree/main/findings

Also, I suppose, committing the plot specification artifacts to spt-data for now. Similar to findings.

spatialprofilingtoolbox/db/scripts/upload_sync_small.py
```diff
@@ -0,0 +1,85 @@
...
+) -> None:
+    for study, study_data in data_per_study.items():
+        _upload_sync_study(study, name, study_data, database_config_file)
+
+
+def main():
+    args = parse_args()
+    if args.name not in APPROVED_NAMES:
+        logger.error(f'{args.name} is not an approved table name.')
+        return
```
jimmymathews, line 76: If we do the argument group or choices thing I recommended, this block's functionality will be handled by argparse.

spatialprofilingtoolbox/db/scripts/upload_sync_small.py
```diff
@@ -0,0 +1,85 @@
+"""Synchronize a small data artifact with the database."""
```
jimmymathews, line 1: Next time, consider using e.g.

```
git mv upload_sync_findings.py upload_sync_small.py
```

to rename the file before making your changes. Git registered the whole file as new, erasing the history.

test/test_data/gnn_plot.json
```diff
@@ -0,0 +1,10 @@
+{
+    "Melanoma intralesional IL2": [
+        "Tumor, Adipocyte or Langerhans cell, Natural killer cell, CD4+ T cell, Nerve, B cell, CD4+/CD8+ T cell, CD4+ regulatory T cell, CD8+ natural killer T cell, CD8+ regulatory T cell, CD8+ T cell, Double negative regulatory T cell, T cell/null phenotype, Natural killer T cell, CD4+ natural killer T cell",
+        "cg-gnn, graph-transformer",
+        "11, 8",
+        "horizontal",
+        "1, Non-responder",
+        "3, Responder"
+    ]
+}
```
jimmymathews, line 10: Can you explain the reasoning here? If it will already be JSON, it seems we can provide annotation/hints that simplify parsing (e.g. key "phenotypes" -> [ "<phenotype 1>", ... ]).

CarlinLiao:
All tests pass except for workflow tests, which complain about `default_study_lookup` not being present. Not sure if it's unrelated or if the change to the first test database broke something.

test/test_data/gnn_plot.json
jimmymathews: The findings format is a list of strings for each study, and after they are inserted into the DB they should probably be regarded as unordered.

Since this usage demands more structure, why not get the uploader to insert the JSON string as a single string value?

jimmymathews: This way the fact that the database table records are in principle unordered will not inhibit robust parsing, and also save the trouble of having to split on commas, etc.

jimmymathews:
I would check that the new system of data-preloaded-images is getting used (in a recent update I moved these images to docker hub, so development doesn't require building them locally).

spatialprofilingtoolbox/db/scripts/upload_sync_small.py
jimmymathews: Gotcha, thanks for doing that.

jimmymathews:
> `test/apiserver/unit_tests/record_counts1.txt` and `build/build_scripts/expected_table_counts.txt` are identical. Is this redundancy necessary?

We will have to check usage, probably the idea was that the test and build scripts would not need to reach into each other's directory contents.

CarlinLiao:
If the data loaded images live remotely, and I've modified one of the data loaded images to have two importance score studies so the plotting has something to compare, would that cause an issue?

jimmymathews:
Yes, this is the scenario where, when this PR is complete, we would update the docker hub images.
I suppose for now you would need to make sure that your development setup is using your locally-built images. I'm not sure what docker does when both remote and local are available, I think it depends on the pull policy.

build/build_scripts/import_test_dataset1.sh
```diff
@@ -10,8 +10,10 @@ nextflow run .
 rm -f .nextflow.log*; rm -rf .nextflow/; rm -f configure.sh; rm -f run.sh; rm -f main.nf; rm -f nextflow.config; rm -rf work/; rm -rf results/
 
 spt graphs upload-importances --config_path=build/build_scripts/.graph.config --importances_csv_path=test/test_data/gnn_importances/1.csv
+spt graphs upload-importances --config_path=build/build_scripts/.graph_transformer.config --importances_csv_path=test/test_data/gnn_importances/1.csv
 
-spt db upload-sync-findings --database-config-file=build/db/.spt_db.config.local test/test_data/findings.json
+spt db upload-sync-small --database-config-file=build/db/.spt_db.config.local findings test/test_data/findings.json
+spt db upload-sync-small --database-config-file=build/db/.spt_db.config.local gnn_plot_configurations test/test_data/gnn_plot.json
```
CarlinLiao, line 16: Both old and new versions of this build script reach into the test directory for data.

jimmymathews:
Tests passed.

PR #322, opened 2024-05-29
Library dependency updates

### LINKED ISSUE #177: API path operation docstrings and examples
The autogenerated documentation for the API path operations supports example parameters and I think possibly examples responses as well. This would be a great addition to improve the usefulness of the docs page.

Also add expanded docstrings for each path operation (currently extremely minimal one-liners).

jimmymathews:
This PR makes changes for consistency with the most recent releases of all dependency libraries. The most significant change was to the squidpy metric tests (current version 1.5.0), as it seems that the neighborhood enrichment and auto-correlation implementations have changed from 1.3.0 to 1.5.0.

All tests pass, though possibly some functionality not directly tested by the test suite is affected, like deployment scripts, uploaders or workflows on current data.

Although this was done in branch `issue177`, this PR does not close #177.


jimmymathews:
I am also provisionally deprecating the graph-plugin-dockerized workflow test, as it takes too long at around 10 minutes. To restore it we will need to fix this (it should not take this long).

jimmymathews:
Before merge the last thing to update is the re-analysis scripts. It is expected that the results related to autocorrelation and neighborhood enrichment will need to be updated.

jimmymathews:
This includes a re-creation of all of those cached computations (cached during application usage) requested by the analysis scripts, and this was a successful test of the recent infrastructure configuration changes.

Issue #339, opened 2024-07-12
Infinite loop checking for completed metric


jimmymathews:
In some cases, the API handler for metric computation requests will repeatedly report that since a job has finished, feature `#` may be complete. This is supposed to happen only when receiving a postgres `NOTIFY` notification that a job completed, but for some reason this can be infinitely repeated.

The real problem here is that for some reason, jobs can be dropped and never completed. I observed a case where this was just 1 out of ~550 (LUAD), and another where it was a few dozen out of ~550. Here dropped means that the job was removed from the queue but no computed value was ever inserted. I don't know what conditions produce this, because in the observed cases a repeat attempt (after deleting the feature's computed values) worked to completion.

jimmymathews:
While waiting to fully diagnose this problem, as possible workaround is to implement some "watchdog" that deletes all values of a feature that is not complete when it seems that all worker instances are idle for sufficient time.

jimmymathews:
For visibility, we may also need to finally implement an additional queue table listing the pending jobs. I avoided this to try to keep the complexity low, but the current system where workers are autonomous and do not communicate or block anything makes debugging challenging.

PR #338, opened 2024-07-10
Vectorize operations on CellDataArrays

CarlinLiao:
To close #337. This should be ready but I've left it as a draft because I won't have access to a machine that can run the Docker test suite until late tonight.

PR #341, opened 2024-07-30
Intermittent bug fixes

jimmymathews:
This PR does two things:

1. Deals with most of #339 by implementing some fallback behaviors in certain error states.
2. Updates an occurrence of feature insertion to fix intermittent test failures noted in the thread of #338.

(1) I was not able to really verify that the check noted in #339 was "infinite", but it is still incorrect and surely related to failing jobs and the complex logic related to requesting that features get computed. This complexity is due mostly to the fact that the counts metric is the only one which is meant to return to clients without any "pending" flag, so the client does not have to poll. I cleaned up this logic a little bit and introduced a 5 minute timeout that clears a feature that seems to have no active jobs and is still incomplete (allowing that it might compute correctly after a new request in the future). I also reduced the number of database connections made by the workers by consolidation.

(2) The `ADIFeaturesUploader` is now only used in one place, but when the schema was changed slightly to use more autoincrementing identifiers, this one usage was not updated, leading to certain errors. This is now updated.

### DIFF
--- a/spatialprofilingtoolbox/db/importance_score_transcriber.py
+++ b/spatialprofilingtoolbox/db/importance_score_transcriber.py
@@ -133,8 +133,6 @@ def _upload(
             get_feature_description("gnn importance score"),
             len(importance_score_set_indexer),
         ),
-        impute_zeros=True,
-        upload_anyway=True,
     ) as feature_uploader:
         for histological_structure, row in df.iterrows():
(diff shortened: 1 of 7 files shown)


jimmymathews:
This also fixes a timing issue that in exceptional circumstances could make the worker containers remain idle, failing to catch the postgres `NOTIFY` signals.

Issue #344, opened 2024-08-02
Counts endpoint meant for polling


jimmymathews:
The endpoint `anonymous-phenotype-counts-fast` is too blocking, implement a variant that, like the other metrics, triggers only a request for computation and immediatley returns the values which have already been computing plus a pending flag.

PR #345, opened 2024-08-02
Add nonblocking counts endpoint

jimmymathews:
Addresses #344.


jimmymathews:
The new endpoint is `phenotype-counts`. The return data type now includes `is_pending`. In the new responses, the "number of cells" field is zero, as it is necessary and present only for backwards compatibility. I plan to deprecate it as soon as the `anonymous-phenotype-counts-fast` endpoint is no longer used.

PR #350, opened 2024-08-26
Supply UMAP position information as virtual slide

jimmymathews:
This is not merged yet, but I released the application images for frontend development purposes.
The functionality unrelated to UMAPs seems unaffected.

You can get the new data just by supplying the sample name "UMAP virtual slide":

```sh
 curl 'http://oncopathtk.org/api/cell-data-binary/?study=Melanoma+intralesional+IL2&sample=UMAP+virtual+sample' > payload.bin
```

View it with:
```sh
tail -c +21 payload.bin | xxd -b -c 20 | less
```

I think we should wait to merge until we have tested it in the frontend.


jimmymathews:
There is also a new flag in the study summary, `has_umap`, to indicate if this is available for the given study.

jimmymathews:
All tests pass except that it seems that there is a test (unrelated to this PR) which is not completely deteriministic (ripley). Might need to search harder for the location to set a random seed, or else relax the test.

PR #360, opened 2024-09-30
API docs

jimmymathews:
#177.


jimmymathews:
This also fixes a couple of performance hits caused by unnecessary computation (`cell_count`).

jimmymathews:
All tests pass.

Issue #368, opened 2024-10-07
Better logs


jimmymathews:
Much of the SPT codebase is good at generating detailed logging, but there are a few specific issues that impair the process of reviewing the logs:

- Too fine-grained reporting by the computation workers. These should perhaps be consolidated into rate-limited summaries, e.g. after a few minutes report the number of jobs taken off the queue, computed, etc.
- A few mistakes in logging logic that create unnecessary warnings (specific example: the feature request / scheduler reporting that "denominator" samples are missing due to mixing up denominator and numerator).

PR #370, opened 2024-10-10
Better logs, and more resilient job queue logic

jimmymathews:
Addresses #368.

Also fixes the job queue logic so that newly created worker containers will first review the queue, without prompting from the message channel it is `LISTEN`ing on in postgres.


jimmymathews:
With scaling (dynamic creation and removal of worker processes), the probably of occasional failed jobs increases (due to differences in memory availability, for example) and we finally need to track these more carefully.
To finish this issue, I am implementing a flag on `quantitative_feature_value_queue` that workers set when they begin computation (rather than pulling off the queue). This may also include a timestamp, so a "watchdog" step can note probably-failed jobs, log a warning, then clean up the corresponding features.

jimmymathews:
The above was completed.

PR #376, opened 2024-10-24
UMAP discrete feature bugfix

jimmymathews:
This PR fixes a bug in UMAP creation where channels in the source datasets are assumed to be continuous-valued and passed through as-is to the reduction. This assumption is normally valid, but sometimes it is helpful to have, provisionally, pseudo-channels that indicate some discrete feature. In these cases we must make sure the discrete features are removed before dimensional reduction is attempted.


jimmymathews:
I also slipped in an update to the versions list that is provided as an API endpoint: to include `spatialprofilingtoolbox` itself.

PR #382, opened 2024-12-02
Top results

jimmymathews:
In the normal development setup,
- local docker compose
- with local services including postgres
- chrome browser in local security mode

it seems that ORCID blocks the sign-in functionality. I tried logging in with my normal credentials directly at orcid.org and it was stopped (probably for a good reason). This worked fine in a normal separate browser session (*where the application in development mode cannot run).

So to test this we might need to create a proper test environment.


jimmymathews:
We can perhaps test the findings-specific functionality (independent of ORCID) by manually bypassing the authentication, just for testing locally.

jimmymathews:
Supposedly fastapi `on_event` has been deprecated:
https://fastapi.tiangolo.com/advanced/events/?h=lifespan#async-context-manager

I updated the on-startup code accordingly.

jimmymathews:
I added a review tool that allows updating the description text and setting the status field.

jimmymathews:
I think the only issue I observed was some of the logic in the dialog:

It seems that by default, after you have submitted one result (by clicking Confirm), the dialog is not dimissed and you can still see the "Submit" button, but the "Confirm" button on the next page is permanently disabled. Probably we should not disable this button, but should rather dismiss the dialog? Or will this automatically happen after redirection back from ORCID?

jimmymathews:
@franciscouzo if you get a chance, see if you can add an additional "description" or "notes" or "background" text field per-finding. The idea is that we would like to have the single-sentence literal description of the finding, but the analyst may also want to provide paragraph-sized background.

jimmymathews:
So we won't need to hard-code the keys.

jimmymathews:
Is there anything remaining for this feature?
It seems complete to me.

franciscouzo:
There are still some minor things I would like to improve, but it's ready to be merged anyway.

Issue #388, opened 2025-02-21
Separate unit tests and integration tests, parallelize tests


jimmymathews:
The current test harness is orchestrated with `make` and docker and so it supports relatively straightforward parallelization of several tests to save time. However several tests have been written in a way that precludes concurrent runs (temporary file artifact name collision, for example). These need to be fixed to allow concurrency and save time in local testing. In provisional runs this reduced the test suite runtime from about 5 minutes to about 1.5 minutes. A start on this work was made in branch `issue359` where the makefile logic was cleaned up somewhat.

There should also be corresponding updates to the maintenance documentation, including some notes on possibly running this test suite in a development cluster rather than locally.

This should also include an initial pass at converting some of the tests into isolated tests of the `spatialprofilingtoolbox` package (rather than whole-application integration tests), and implementing a github action to run this test suite (this should complete in just a few seconds).

Issue #359, opened 2024-09-29
Dependency updates


jimmymathews:
A number of dependencies have updates, some quite major. For this issue, update the following if possible and refactor as necessary:

1. Python 3.11 => 3.13. Version 3.13 is supposed to be released October 1, but I think I read that this might be pushed back a week or more.
2. Postgres 14 => 17. Version 17 was very recently released. Previously I recall some minor reason that it was beneficial to hold back our postgres version, I think to take advantage of the official Postgres docker images for our test containers. Try to resolve this and upgrade everything to v17.
3. `pip list --outdated` reports:
```txt
Package       Version  Latest
------------- -------- ------
attrs         23.2.0   24.2.0
bokeh         3.4.1    3.6.0
boto3         1.35.23  1.35.29
botocore      1.35.23  1.35.29
fastapi       0.111.0  0.115.0
fsspec        2023.6.0 2024.9.0
h5py          3.10.0   3.12.1
llvmlite      0.42.0   0.43.0
numba         0.59.1   0.60.0
numpy         1.26.4   2.1.1
pandas        2.2.2    2.2.3
Pillow        9.5.0    10.4.0
pip           24.0     24.2
psycopg       3.1.19   3.2.3
pydantic      2.7.2    2.9.2
pydantic_core 2.18.3   2.24.0
s3fs          0.6.0    2024.9.0
scikit-learn  1.5.0    1.5.2
scipy         1.13.1   1.14.1
secure        0.3.0    1.0.0
setuptools    65.5.0   75.1.0
squidpy       1.5.0    1.6.1
starlette     0.37.2   0.39.2
tables        3.9.2    3.10.1
tqdm          4.66.4   4.66.5
uvicorn       0.30.0   0.31.0
```

PR #389, opened 2025-02-21
Version management

jimmymathews:
Addresses #359.

This PR updates the codebase to the most recent possible versions of all upstream dependencies, and introduces an easier method to do such updates in the future. This includes updating for Python 3.13 and PostgresQL 17.2.

`pyproject.toml` remains our place to declare the unversioned Python dependencies. Version specifications that represent the most up-to-date tested environment are committed to version control in `requirements.txt` (and `requirements.apiserver.txt` and `requirements.ondemand.txt` for the application container images' dependencies).

`requirements.txt` is determined using `uv` for a more efficient build cycle than we would have with plain `pip`. To reduce external dependencies, however, `uv` is not used for actual installation into the container images.

Also, a few adjustments were made to the `make test` process, mainly how logs are saved and displayed as well as removing some name collisions (filenames, container names) between tests, in preparation for #388 .

Notes about specific packages:
- `spatialdata`. This package is held back to version `0.2.5` (current is `0.3.0`) due to lack of python 3.13 support. This is fine for now as we do not directly use it, but this is relevant as we would like to eventually attempt at least data exports into spatialdata-compliant formats.
- `psycopg`. Handling of notifications was slightly altered, which required a change to our usage of postgres notify.
- `dask`. Some API changes in this indirect dependency are handled poorly in some libraries, forcing us to import it as though a direct dependency and set some settings at the moment. In the future we should try to remove this special treatment.
- `secure`. The API changed requiring some minor updates, and the headers that are generated in our current configuration are different from before, so the new values will probably need to be checked for adequacy.

Before this PR is ready to be merged, there are some fixes that should be done first that reduce unnecessary cross-module imports in the apiserver and ondemand modules. The ondemand `providers` should have their feature-computation-scheduling-related functionality factored out and moved to the apiserver, so that the apiserver container image does not require certain slow imports of unused packages (e.g. squidpy). The container image build should also include a step that pre-creates whatever cache it is that makes python imports faster the second time (the container startup is likely unnecessarily slow).


jimmymathews:
I did the last item described above. I refactored the queue management stuff so that everything required for insertion of jobs, as initiated within `apiserver`, is contained in `apiserver.request_scheduling`, and everything required for popping jobs, as initiated by `ondemand` computation workers, is contained in `ondemand`. The computation worker classes (e.g. `CountsComputer`) now are very focused on their specific computation. Moreover the original aim of reducing the library dependencies of the `apiserver` container was achieved (e.g. `squidpy` is now only used by `SquidpyComputer` in the `ondemand` container).

I also identified a bug that was caused by the usage of an ORM in the newest API endpoints. The setup functions to create the database schema for "findings" were crashing due to existing tables, I believe causing the effective number of uvicorn workers (parallel processes for the API server) to be reduced. I fixed this by deprecated the ORM usage entirely, using plain SQL to add `finding` to the data model (as is the pattern for our other tables) and to insert/select. Consequently the library dependency is reduced further, omitting `sqlmodel` and `alembic`. Our test for the findings functionality was not very comprehensive, so I also added a new test that checks that insertion/selection (post/get) works to round-trip a record.

jimmymathews:
Regarding the dependency-minimization effort: The compressed `apiserver` image is only 60% of the size of the previous compressed image. The `ondemand` is also smaller but only slightly.

Issue #387, opened 2025-02-21
Race condition for computed feature specifications


jimmymathews:
There is a rare race condition in computation of specified features, due to the way that each feature specification is spread over multiple tables (namely `feature_specifier` and `feature_specification`). This design was chosen for flexibility, rather than dealing with schema mutation to `feature_specification` to accomodate different feature types. It should only take a few milliseconds to insert, but if the same specification's insertion is requested simultaneously from two sources, it's likely that a duplicate will occur, and I have observed such duplication in testing.

This can be fixed easily by adding a new table like `feature_specification_hash(feature_specification, hash)` where the `hash` is a string serialization of the specifiers and `hash` has a unique constraint. We would then attempt insertion into this table in a transaction containing the feature specification insertions, and on conflict rollback the transaction.

PR #391, opened 2025-02-27
Add unique identity for features

jimmymathews:
Addresses #387.


jimmymathews:
This was tested with recreation of several metrics using `run_all.py`, no errors were observed.

PR #392, opened 2025-02-28
Test harness update

jimmymathews:
Meant to address #388.


jimmymathews:
Parallel `make` now works with speedup from about 4.5 minutes down to 1.5 minutes with ~10 cores.

jimmymathews:
The separation of unit tests is deferred, the parallelization is complete.

PR #400, opened 2025-04-03
Add some database-wide convenience annotations

jimmymathews:
For a final presentation layer, it is useful to have some annotations of channels like convenience groupings, aliases, and preferred colors.
This PR adds a CLI command to upload such annotations, and an API endpoint which provides these to client applications.


jimmymathews:
Tests pass, including a new test for these annotations.

Issue #401, opened 2025-04-07
Implement batch operation for some API endpoints


jimmymathews:
For the `phenotype-counts` endpoint, there are usages where several (~100-200) requests may be made at once for a single display/functionality. A lot of time is wasted in making networks connections, so this takes a couple minutes.
Implement a batch version of this endpoint that only requires one request.

Issue #399, opened 2025-04-01
Channels not validated


jimmymathews:
I think that channel names are not validated for the given study when certain metrics are requested. Channels occurring  somewhere but not occurring in the requested study are not flagged, and the metrics computation proceeds to a failure state.

jimmymathews:
What was happening is that channels are only validated in the sense of occurrence or not across all datasets. They needed to be validated in a study-specific way.

PR #402, opened 2025-04-07
Add batch counts endpoint

jimmymathews:
Addresses #401.


jimmymathews:
Also addresses #399 .

Issue #419, opened 2025-05-30
Offload big tables to S3


jimmymathews:
# Overview
This is a plan to reduce unnecessary long-term memory usage in the Postgres database, reduce ETL resource usage (time/bandwidth), and simplify backup/restore and some CI actions.

# Purpose

- Reduce PostgresQL table memory usage from about 191 GB to about 1 GB (plus 4 GB in new S3 buckets), enabling a much more robust backup process and automatic management of database state.
- Reduce the number of uploads/downloads of the big payloads (network bandwidth usage) during the dataset upload process.
- Eliminate some unnecessary indexes, range definitions, and queries associated with oversized tables.
- Simplify parts of the codebase involving these large tables and some parts involving the feature matrix payloads.

# Plan

In the diagram, before is above and after is below.

![Image](https://github.com/user-attachments/assets/49e06ec5-56b0-4b66-ace8-4d0a41b4b7b2)

1. Make the ETL process (phase `spt workflow initialize`, mainly `skimmer.py`) stop uploading data to big tables:
    - `shape_file`
    - `expression_quantification`
    - `histological_structure_identification`
    - (maybe also `histological_structure`)
2. Make the ETL process (phase `spt ondemand assess-recreate-cache`) stop downloading data from those big tables.
3. Merge these two phases (skimmer.py and recreate cache) so that the source files for a sample (e.g. `22.csv`), either regenerated locally from upstream sources or downloaded from S3):
    - are loaded into local memory.
    - Then directly create the things currently created by:
        a. `BinaryFeaturePositionsCacheManager`
        b. `UMAPCacheManager`
        c. `CompressedPayloadsCacheManager`
        d. `IntensitiesCacheManager`
    - Results uploaded, no longer to table `ondemand_studies_index`, but instead directly to S3.
        - Top-level path components can look like:
            `feature_matrix`
            `cell_data_brotli`
            `expressions_index`
            `feature_matrix with intensities`
            `centroids`
        - Possibly deprecate the `expressions_index` or improve it to be better documented as metadata for the feature matrices. This could involve updating the `scstudies` "ADI" schema in order to apply to the syntax of data artifacts more general than tables. That is, something in addition to `tables.tsv` and `fields.tsv` which documents rows, columns, and values for feature matrices.
        - Filenames can be (sanitized/normalized) sample names, e.g.:
            - `h_p_64_1`
            - `h_p_65_0`
            - ...
[...]

PR #420, opened 2025-06-17
Remove dependency on large table query

jimmymathews:
After some of the cleanup in #419 was performed, the live application was relying on some cached values that would no longer be accessible directly from the database.
This quick fix updates the usage (`get_number_cells`) to use the compressed file-based sample data instead.


jimmymathews:
Some tests are failing (partly related to updated libraries, partly related to the need from refactoring access to the feature matrices in some test scenarios), but these are not critical.

PR #424, opened 2025-07-24
Add Content-Disposition header to /sqlite/

franciscouzo:
This is a proof of concept, I wasn't able to test the change, since I ran into some issues with the docker container.


jimmymathews:
For me using Safari, Chrome, and Firefox this already works due to the `<a>` tag's `download` attribute.
It seems that a benefit of adding this server-side header is that future occurrences of links will have the correct filename even if the wrapping `<a>` tag doesn't specify it.

franciscouzo:
The download attribute only works on same-origin links (such as on the production website), when trying to download the link on development, the downloaded filename will be something generic set by the browser.

jimmymathews:
(I also made some minor unrelated updates to make some experimental artifacts up-to-date, those `analysis_replication/...` files).

Issue #404, opened 2025-04-11
Create pre-subsampled intensity data bundles


jimmymathews:
For planned work on new threshold-adjustment features in the UI, we will need the server to provide a batch of cell intensity data across all samples in a study, subsampled for performance/memory reasons.

This should probably always send back roughly the same number of cells regardless of study size, to normalize the client-side handling process. Around 500k-1m is probably doable.

It will be helpful for this to be a single payload. It will contain raw data but should also contain some metadata, so perhaps it can be a zip of one large binary file (similar format to our cell intensity payloads), and a JSON file with easily-destructed metadata including offsets per sample and channel names.

jimmymathews:
In the protobufs documentation they recommend against protobuffers for more than a few megabytes payloads (https://protobuf.dev/overview/).
@franciscouzo what do you think about using the ASCII file separator (decimal 28) to delimit a JSON file with metadata and the main binary portion. It's a little old-fashioned but accurate in this case.

jimmymathews:
I'm not sure if ASCII 28 is allowable in JSON but in any case we won't use it in this particular JSON payload.

jimmymathews:
I'm documenting the metadata format here in [cells.md](https://github.com/nadeemlab/SPT/blob/issue404/docs/cells.md) .

jimmymathews:
I just remembered that the thresholds will often be specimen-specific, so the thresholds need to go inside the per-sample section.

Since I want to make sure the format advertises a single channel order I think I'll do it like this:

```json
{
  "subsample_counts": [
    {
      "specimen": "ABC_001",
      "count": 2500,
      "thresholds": {"CD3": 0.6, "CD4": 0.4, ...}
    },
    {
      "specimen": "ABC_002",
      "count": 2000,
      "thresholds": {"CD3": 0.3, "CD4": 0.3, ...}
    },
    ...
  ],
  "channel_order": ["CD3", "CD4", ...]
  ]
}
```

There is a problem though: this creates a burden on the threshold adjustment UI to somehow normalize the values across samples.

Do you think we should just make the backend pre-normalize (with a multiplicative scale) so that the threshold is always 0.5?

jimmymathews:
Although it is annoying, I think the final output of the threshold adjustment probably should be specimen-specific, because the original intensity values are not generally harmonized between samples. We want the output artifact to be usable for thresholding these original intensity values.

[...]

PR #425, opened 2025-08-11
Create aggregated downsampling

jimmymathews:
Addresses #404 .

- Unfortunately the `cache_subsample` script is largely copied from the interactive uploader script, because I wanted to reuse the functionality of selecting the database interactively but it was not yet factored out. In the future this can be cleaned up.
- The basic functionality seems to be all here, I just need to add a roundtrip test.


jimmymathews:
This has been deployed to production, you can see docs for the new endpoint:
https://oncopathtk.org/api/redoc#operation/get_cell_data_binary_intensity_whole_study_subsample_cell_data_binary_intensity_whole_study_subsample__get

You can see the pattern of extraction of the payload here:
[downsampled_retrieval.sh](https://github.com/nadeemlab/SPT/blob/issue404/test/apiserver/module_tests/downsampled_retrieval.sh)

10 out of the 12 datasets have this payload precomputed.

franciscouzo:
Having the counts info and the binary data in the same file makes parsing it pretty cumbersome, I'm trying to avoid making copies of the buffer to prevent using too much memory and slowdowns, but it's complicated, I would have to refactor a lot of code to use` DataView`s and passing `byteOffset` to `Uint8Array`.

It would be way simpler to have one endpoint return the json, and another the binary data

jimmymathews:
Sure, I'll just make a separate endoint for the metadata.
As we discussed, the benefit of combining them was to provide evidence to the API user that the two pieces are not out of sync, so we will lose that, but it's OK.

franciscouzo:
I don't think that should be an issue, the intensity data won't change, and we are already doing it that way with the per-sample intensities

PR #444, opened 2026-04-08
Make tests pass and do refactoring (deprecations etc.) for ETL consolidation

jimmymathews:
This PR does a number of updates to the new #419 work. Many refactorings were enabled by the ETL consolidation, resulting in deprecation of several now-obsolete modules and around 2000 lines of code. This will help with ongoing maintenance, debuggability, etc.

I also made sure the tests are passing now. The exception here is a number of tests related to the GNN functionality, because #419 has called for deprecation of study-global cell-level identifiers. So we will need to update the GNN stuff to record explicitly the sample alongside the cell IDs it uses.

Note that the S3 option as the backing store is not yet tested, and it will likely not work as-is without some tweaks.

- Deprecates several modules related to intermediate data storage/transformation which are no longer used after merging the ETL steps into a more local operation:
  - `assess-recreate-cache` script
  - `cache_assessment`
  - `cache_pulling`
  - `sparse_matrix_puller`
  - `structure_centroids_puller`
- Deprecates CLI `count-cells`, as this now runs automatically after the main ETL
- Adds blob retrieval and key existence check to new cache store abstraction
- Uses the new cache store in:
  - cells data accessor
  - feature names saving/retrieval
  - `representative_subsample`, the module for doing whole-study subsamples
  - the part of `generic_job_computer` that pre-loads cell data for a generic metrics computation
  - UMAP creation
- Gathers all transformation into and out of the binary format into `CompressedMatrixHandling`, including from `CellsAccess`
- For `FeatureMatrixRetrieval` (renamed from `FeatureMatrixExtractor`):
  - simplifies implementation
  - expands usage of this class by other modules, replacing duplicated functionality
  - refactoring to use the new cache store as the source of binary payloads
- Fixes a bug in continuous dataframe parsing where the first 4 bytes of each row were accidentally parsed as channel data (these contain a cell ID), and corresponding test data artifact update
- Simplifies the threshold determination in representative subsampling
- Do all the deprecations in `cell_manifests` parsing that are possible now that a separate parser is implemented, e.g. the shape file stuff, and modularize the previously long/monolithic nested loops for processing each sample/cell
- Deprecates the "expressions index" with its convoluted associations between some identifiers and feature names
- Deprecates `RangeDefinition`
- Deprecates "autocomputed" metrics, only specifically requested metrics will now be computed
- Deprecates `fetch_one_or_else`, previously used as a convenience SQL operation but no longer used
- Addresses several occurrences of `psycopg` typing complaints about string literals as queries
- Updates a few tests that were failing and deprecate a few obsolete tests
- Updates example data artifact for prior change (surivival-type data, unrelated to new ETL)


jimmymathews:
I tested this on the example dataset in `dataset_curation/datasets`, using a local application instance to poke around and search for upload/processing errors. Although it is only local import, so the comparison isn't the same as for the production use-case, I did observe about 4x speedup compared with previous non-consolidated ETL. It takes about 4 minutes for the ~100k dataset, about 2 minutes of which is just UMAP generation and so would not scale up with the dataset size. Setting aside metadata stuff, the main loop was quite fast, a few seconds per sample.

(One new behavior: I used uniform scaling to the UMAP intensity channels to fix what had been an erroneous clipping on all channels, which brings it into conformity with the uniform scaling used on intensity channels generally before import.)

PR #448, opened 2026-05-26
Cell atlas

jimmymathews:
This is my proposal for finishing the backend portion of [smprofiler-web#224](https://github.com/nadeemlab/smprofiler-web/issues/224). (For @wooferclaw ).

1. Add the main new code (`train_atlas_models.py`) to the smprofiler codebase, probably also breaking it down, i.e. modularizing it by bringing together related functions and placing function/class definitions in a logical order. At the minimum, a portion (separate file) that is ordinary library code, and a portion that is the CLI under `smprofiler/<module>/scripts/...`.
2. Manually rewrite the inline documentation to make sure it is really needed and accurate (it looks like some automatic tools were used to generate this). I can be more specific about this if that would help.
3. In the future the HGNC normalization may be helpful, but for the current purpose it is not needed because we have a complete manual mapping between the channel names as used in our datasets and the channel names appearing in the cell atlas. So this functionality should perhaps be split off as a separate module (file in the `smprofiler` codebase).
4. Refactor the initial portion of the training to accept its input from the Parquet-formatted table file generated by [this aggregation](https://github.com/nadeemlab/smprofiler-data/tree/cell_atlas/cell_atlas_preprocessing) script in `smprofiler-data`. (Eventually I'll probably bring this functionality into `smprofiler` as well, for convenience.)
5. Include a tiny test of the training using a tiny subset of the real atlas data.
6. Maybe deprecate the ability to load the channel annotations from file. That way we only get it from the API, to help make sure we don't use mismatched source code vs. live versions.
7. Consider deprecating or justifying some of the additional "utility" functions like `_silence_fd`. Maybe we can just use `print`. There is also a logging pattern used throughout the `smprofiler` codebase:
```py
from smprofiler.standalone_utilities.log_formats import colorized_logger
logger = colorized_logger(__name__)
...
logger.info('...')
logger.error('...')
logger.warning('...')
```
8. Ensure that the only model architectures that are considered provide a non-trivial variance/standard deviation estimation method during inference. By non-trivial I mean that it should actually depend on the input `x` provided for inference, not just the global standard deviation of the predictions over all the training data (or similar). The only architecture I'm sure provides this is the Bayesian ridge and Gaussian Process regression. However if we can make it work well, a bootstrapping procedure may work here for an arbitrary architecture, provided we do some kind of weighting of the training data samples reflecting similarity with the provided input `x`.
9. Include some basic functionality for storing the models once trained. I think a new database table would be fine for this, as the models are quite small, and we could easily store important metadata this way. This could allow, for example, multiple models trained at different times (older/newer versions). The new table could be along the lines of:
```sql
CREATE TABLE(
    onnx_model BYTEA,
    study VARCHAR,
    FOREIGN KEY (channel_predicted) references chemical_species(id),
    training_time_minutes NUMERIC,
    created TIMESTAMP WITH TIMEZONE,
    size_bytes INTEGER,
    architecture_type VARCHAR
)
```
10. Add some separate entrypoints into the functionality that are documented for function usage, not implementation. Things like `get_model` (or similar-named) API endpoint, a model usage (inference) snippet (Python, to be tested in one of the small tests). A small JS usage snippet would also be useful.


wooferclaw:
@jimmymathews Hi, Jimmy! I have pushed the changes that cover most of the issues you have mentioned above - can you please provide some feedback?

jimmymathews:
Thanks! Yes I will take a look.

jimmymathews:
I am working through this, looks good so far.
One thing is I want to make sure that the z-scores for a given functional marker prediction are the primary output, i.e. the standard deviation is accounted for at inference time using our generated ONNX models. The onnx runtime docs are a little vague on this point, but I found one helpful note specifically about the `predict_std=True` option in the case of one of our sklearn model types: https://onnx.ai/sklearn-onnx/auto_examples/plot_gpr.html#return-std-true

wooferclaw:
implemented some STD experiments in https://github.com/nadeemlab/smprofiler/tree/std-deviation-experimental another branch - please take a look

jimmymathews:
I read through the experiments in that branch. Sorry it looks like I duplicated in my PR (`edits_mlops`) some of what you have there, related to the final z-score computation.

In `_append_bayesian_ridge_std` there is a detailed usage of the ONNX internals, appending ONNX graph nodes and operations. I was hoping we would be able to delegate such low-level operations to a library, i.e. `skl2onnx` in this case, and allow ourselves to treat the ONNX models as an efficient/slim black box. In the usage in my branch `edits_mlops` I think I was able to achieve this while still outputting the stddev.

On the other hand, if we *can* reliably make use of lower-level ONNX API: I did consider trying to make the ONNX model itself perform the final operation of forming the z-score / standard deviate as the primary output, assuming we can also make sure the measured functional column is also always included as an input. That way model users would have an even easier pathway to using the model. They would not have to follow our specific snippet for post-processing the outputs.

Regardless we would still need to track down the "bug" of totally inconsistent stddev between the sklearn model and ONNX conversion.

Also let me re-mention a suggestion I made earlier: For any architecture (including those random forest ones), there is a possible mechanism of determinining meaningful per-sample standard deviation using bootstrapping: Embed a random subsample of the training set in the model, and at inference time use a distance-to-training-points kernel as weights for estimating the "predictive distribution" (then we can determine any moments we want from this distribution). If we cannot resolve the stddev reproducibility issues we might have to fall back to such bootstrapping.

jimmymathews:
I will try to run the training on the stddev experiments branch, to see if the concordance between sklearn/ONNX is any better.

jimmymathews:
Regarding `_append_bayesian_ridge_std`: I just noticed the note claiming the `skl2onnx`'s implementation of the BayesianRidge `return_std=True` option is inaccurate, is there some reference for this? That would explain the trouble I observed. And is there a reference for the alternative formula implemented in `_append_bayesian_ridge_std`?

jimmymathews:
If there are errors in `skl2onnx` that are this simple, maybe we can just fix them and open a PR.

jimmymathews:
It looks like `skl2onnx`'s implementation is here:

https://github.com/onnx/sklearn-onnx/blob/main/skl2onnx/operator_converters/linear_regressor.py#L154-L191

wooferclaw:
@jimmymathews i have updated https://github.com/nadeemlab/smprofiler/tree/std-deviation-experimental according to your feedback

wooferclaw:
Please check the updated version

PR #449, opened 2026-06-10
Fix stale database connections

jimmymathews:
When migrating to the cache store, the main use was initially during ETL, so it was not noticed that each cache store object holds a database connection until end of process.

In the context of the API server that is a new connection for each request (for cache-store data), that is never closed.


jimmymathews:
Originally the point of managing the database connection manually was to ensure that during ETL we wouldn't create a new connection for every blob. That resulted in much faster runs.

For the API server the async architecture prevents us from re-using the same connection across multiple requests. I believe there is some kind of pool we could use that would work in an async manner, but for now the quick fix is to close the connections right after they are used.

jimmymathews:
In order to run tests I am dealing with some minor unrelated changes to nextflow config syntax for env variables.

jimmymathews:
The tests pass (except for some expected failures, cell-level identifiers stuff from old changes).

